PR #5, opened 2015-08-18
add startup script

rudigiesler:
Add a startup script that will run the api (and add it to setup.py)


rudigiesler:
The coverage decrease is because I'm not sure how, or if it's useful, to test the `main` function

rudigiesler:
Ready for review

hodgestar:
:+1:

PR #7, opened 2015-08-26
Twisted parent and child services

rudigiesler:
Have the junebug API run as a twisted multiservice, and add the machinery to launch transports as child services of the API.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -133,9 +133,14 @@ class JunebugApi(object):
                 },
             },
         }))
+    @inlineCallbacks
     def modify_channel(self, request, body, channel_id):
         '''Mondify the channel configuration'''
-        raise NotImplementedError()
+        channel = yield Channel.from_id(
+            self.redis_config, self.amqp_config, channel_id, self.service)
+        returnValue(response(
+            request, 'channel updated', (yield channel.update(body))))
+
 
     @app.route('/channels/<string:channel_id>', methods=['DELETE'])
     def delete_channel(self, request, channel_id):
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -1,33 +1,40 @@
 from klein import Klein
 import logging
 from werkzeug.exceptions import HTTPException
+from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.web import http
 
-from junebug.validate import body_schema, validate
+from junebug.channel import Channel, ChannelNotFound, InvalidChannelType
+from junebug.error import JunebugError
 from junebug.utils import json_body, response
+from junebug.validate import body_schema, validate
 
 logging = logging.getLogger(__name__)
 
 
-class ApiUsageError(Exception):
+class ApiUsageError(JunebugError):
     '''Exception that is raised whenever the API is used incorrectly.
     Used for incorrect requests and invalid data.'''
+    name = 'ApiUsageError'
+    description = 'api usage error'
+    code = http.BAD_REQUEST
 
 
 class JunebugApi(object):
     app = Klein()
 
-    def __init__(self, service):
+    def __init__(self, service, redis_config):
         self.service = service
+        self.redis_config = redis_config
 
-    @app.handle_errors(ApiUsageError)
-    def usage_error(self, request, failure):
-        return response(request, 'api usage error', {
+    @app.handle_errors(JunebugError)
+    def generic_junebug_error(self, request, failure):
+        return response(request, failure.value.description, {
             'errors': [{
-                'type': 'ApiUsageError',
+                'type': failure.value.name,
                 'message': failure.getErrorMessage(),
                 }]
-            }, code=http.BAD_REQUEST)
+            }, code=failure.value.code)
 
     @app.handle_errors(HTTPException)
     def http_error(self, request, failure):
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -17,6 +17,9 @@ class ApiUsageError(Exception):
 class JunebugApi(object):
     app = Klein()
 
+    def __init__(self, service):
+        self.service = service
+
     @app.handle_errors(ApiUsageError)
     def usage_error(self, request, failure):
         return response(request, 'api usage error', {
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -92,8 +92,9 @@ class JunebugApi(object):
     def create_channel(self, request, body):
         '''Create a channel'''
         channel = Channel(
-            self.redis_config, self.amqp_config, body, parent=self.service)
+            self.redis_config, self.amqp_config, body)
         yield channel.save()
+        yield channel.start(self.service)
         returnValue(response(
             request, 'channel created', (yield channel.status())))
 
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -99,9 +99,13 @@ class JunebugApi(object):
             request, 'channel created', (yield channel.status())))
 
     @app.route('/channels/<string:channel_id>', methods=['GET'])
+    @inlineCallbacks
     def get_channel(self, request, channel_id):
         '''Return the channel configuration and a nested status object'''
-        raise NotImplementedError()
+        channel = yield Channel.from_id(
+            self.redis_config, self.amqp_config, channel_id, self.service)
+        returnValue(response(
+            request, 'channel found', (yield channel.status())))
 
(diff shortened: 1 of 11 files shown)


rudigiesler:
Ready for review.

junebug/service.py
```diff
@@ -0,0 +1,32 @@
...
+    def startService(self):
+        '''Starts the HTTP server, and returns the port object that the server
+        is listening on'''
+        self.api = JunebugApi(
+            self, self.redis_config, self.amqp_config)
+        self._port = reactor.listenTCP(
+            self.port, Site(self.api.app.resource()),
+            interface=self.interface)
+        log.msg('Junebug is listening on %s:%s' % (self.interface, self.port))
+        return self._port
```
hodgestar: We should probably not return anything here because `IService` doesn't mention that -- https://twistedmatrix.com/documents/15.3.0/api/twisted.application.service.IService.html

junebug/tests/helpers.py
```diff
@@ -0,0 +1,96 @@
...
+        '''Creates an existing channel given the channel id'''
+        return Channel.from_id(redis._config, {}, id, service)
+
+    @inlineCallbacks
+    def get_redis(self):
+        '''Creates and returns a redis manager'''
+        if hasattr(self, 'redis'):
+            returnValue(self.redis)
+        persistencehelper = PersistenceHelper()
+        yield persistencehelper.setup()
```
hodgestar: We should add `persistencehelper.cleanup` as a clean up function.

junebug/channel.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(
+            self, redis_config, amqp_config, properties, id=None):
+        '''Creates a new channel. ``redis_config`` is the redis config, from
+        which a sub manager is created using the channel id. If the channel id
+        is not supplied, a UUID one is generated. Call ``save`` to save the
+        channel data. It can be started using the ``start`` function.'''
+        self._properties, self.id = properties, id
+        if self.id is None:
+            self.id = str(uuid.uuid4())
+        self._redis_base = RedisManager.from_config(redis_config)
```
hodgestar: Ask discussed at the office, this needs to be a TxRedisManager (just noting it here).

rudigiesler:
The last commit is kind of large. Basically, it changes the channel implementation to take a redis manager instead of a redis config. This allows us to cache the redis manager in the API, instead of creating a new one for each request. It also makes sure that TxRedisManager is used everywhere.

junebug/tests/helpers.py
rudigiesler: Added

junebug/channel.py
```diff
@@ -0,0 +1,149 @@
...
+            del self.transport_worker
+
+    @inlineCallbacks
+    def save(self):
+        '''Saves the channel data into redis.'''
+        properties = json.dumps(self._properties)
+        channel_redis = yield self.redis.sub_manager(self.id)
+        yield channel_redis.set('properties', properties)
+        yield self.redis.sadd('channels', self.id)
+        yield channel_redis.close_manager()
```
rudigiesler: This is one thing that I'm not sure about with the new way of doing this. Do we need to close the `sub_manager` here? Or does it get closed when we close the root manager?

junebug/api.py
```diff
@@ -77,14 +94,29 @@ def get_channel_list(self, request):
...
+    @inlineCallbacks
     def create_channel(self, request, body):
         '''Create a channel'''
-        raise NotImplementedError()
+        redis = yield self._get_redis_manager()
+        channel = Channel(
+            redis, self.amqp_config, body)
+        yield channel.save()
+        yield channel.start(self.service)
+        yield redis.close_manager()
```
justinvdm: If we close the manager each request, what happens on the next request that needs the manager?

junebug/service.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, interface, port, redis_config, amqp_config):
+        super(JunebugService, self).__init__()
+        self.interface = interface
+        self.port = port
+        self.redis_config = redis_config
+        self.amqp_config = amqp_config
+
+    def startService(self):
+        '''Starts the HTTP server, and returns the port object that the server
+        is listening on'''
```
justinvdm: Not too sure on this, but we maybe do to do the `super()` dance here? Similarly for `stopService()`

junebug/command_line.py
```diff
@@ -27,6 +26,42 @@ def parse_arguments(args):
...
+        type=int,
+        help='The port of the amqp endpoint. Defaults to "5672"')
+    parser.add_argument(
+        '--amqp-user', '-amqpu', dest='amqp_user', default='guest',
+        type=str,
+        help='The username to use for the amqp auth. Defaults to "guest"')
+    parser.add_argument(
+        '--amqp-password', '-amqppass', dest='amqp_pass', default='guest',
+        type=str,
+        help='The password to use for the amqp auth. Defaults to "guest"')
```
justinvdm: Did we decide to go with command line args instead of a config file? Seems like we might end up with pretty complicated commands. I guess maybe it isn't too important though.

junebug/channel.py
```diff
@@ -0,0 +1,149 @@
...
+    '''Raised when an invalid channel type is specified'''
+    name = 'InvalidChannelType',
+    description = 'invalid channel type'
+    code = http.BAD_REQUEST
+
+
+transports = {
+    'telnet': 'vumi.transports.telnet.TelnetServerTransport',
+    'xmpp': 'vumi.transports.xmpp.XMPPTransport',
+}
```
justinvdm: Should this maybe be configurable? We seem to be adding new transports as separate python packages, so we'd need to include them as dependencies if we wanted to put them in this list

junebug/command_line.py
rudigiesler: I was wondering about this. It started with command line args, so I just continued it. Maybe we could do both? I just thought it might be strange to create and specify a config file just to change the port that junebug is running. Maybe we could have both (in that you have a `--config-file=..`, or you can specify via the command line)

junebug/api.py
rudigiesler: Yep I realised this. It was magically working because I wasn't caching the redis manager properly in `_get_redis_manager`, so it was just creating a new manager for every request.

junebug/channel.py
rudigiesler: I'm not sure what you mean by configurable? My idea was to add them individually like this, and them document them in the Junebug documentation. It means that you don't have to have `vumi.transports....` in your API request, and also documenting them separately means that we can document what each `config` thing it needs is.

I see that adding them as dependancies might be an issue, but if we're planning to release this as a debian package, wouldn't we need to include them all anyway?

junebug/command_line.py
justinvdm: If there are use cases for both, then both sounds good.

junebug/channel.py
justinvdm: Update from RL discussion: configurable as in the ability to specify transport name to python path mapping in config file, using the dict above as the default. Not super important, so we are leaving it as is for now.

junebug/command_line.py
justinvdm: Maybe something for a separate PR?

junebug/service.py
```diff
@@ -0,0 +1,36 @@
...
+        super(JunebugService, self).__init__()
+        self.interface = interface
+        self.port = port
+        self.redis_config = redis_config
+        self.amqp_config = amqp_config
+
+    @inlineCallbacks
+    def startService(self):
+        '''Starts the HTTP server, and returns the port object that the server
+        is listening on'''
```
justinvdm: Not too sure on this, but I think we might need to do the `super()` thing here and for `stopService()`.

justinvdm:
Left a few more comments, but looks good. Noticed there are `Channel` methods still calling `close_manager`, but I'm probably reviewing this before its ready for re-review.

justinvdm:
Err, re those extra `close_manager` calls, just realised I was looking at an earlier commit, I see those calls are now gone, sorry about that.

junebug/command_line.py
rudigiesler: Yep, I think it would be good to add a "Get config from file" PR. Issue: https://github.com/praekelt/junebug/issues/8

rudigiesler:
I think it's ready for another review.

justinvdm:
:+1:

PR #11, opened 2015-09-01
Support listing of channels

justinvdm:


justinvdm:
Ready for review.

junebug/tests/test_api.py
```diff
@@ -57,14 +57,26 @@ def test_http_error(self):
...
-                    'type': 'NotImplementedError',
-                    }]
-                })
+        yield self.assert_response(resp, http.OK, 'channels listed', [])
+
+        yield Channel(redis, {}, config, u'test-channel-1').save()
+
+        resp = yield self.get('/channels')
+        yield self.assert_response(resp, http.OK, 'channels listed', [
+            u'test-channel-1',
```
hodgestar: The docs at http://junebug.readthedocs.org/en/latest/http_api.html#get--channels don't specify what to return, but since real ids are uuids, it might be nicer to return some information about the channel in addition to the ids? So a list of dictionaries rather than just a list of ids. Thoughts?

rudigiesler:
:+1:

junebug/tests/test_api.py
rudigiesler: To do this it would be an extra redis request per channel, which could be quite costly. On the other hand, if for most use cases we're going to have to get all the data for all the channels, then it's better to send it all on one request. How do we see this being used?

hodgestar: Mostly my thought was that a list of UUIDs was not that useful, and that it would be nice to add the names at least. If we can add more data easily once we have the names, then I'm +1 on adding that (or even just including the complete data if we have it).

justinvdm: Hmm. Names would be nice, also had that thought when I started on this PR. UUIDS are pretty useful by themselves though, they can serve as a starting point to find out more information about each channel via further api calls. I think its common for `GET`s to a collection URI to return a list of URIs of its members, rather than ids or the full object for each member (sounds like it could get pretty large pretty quickly, but maybe not). Returning just a few fields, like the name and id seems a bit weird to me, get the feeling we'd eventually end up returning the entire object for each member at some point if we did that.

That all said, I have a slight preference for UUIDS or URIs, but happy to go with returning the full objects if people prefer.

hodgestar: Since we have different intuitions about this, perhaps lets leave things as
they are and make a note to check with Nic and Mark what they think (since
they'll likely be the first users other than us)?

justinvdm: +1, sounds good

hodgestar:
:+1:

PR #12, opened 2015-09-02
Add /s for collection api endpoints

justinvdm:
To be more RESTful, it would be nice to have collection api endpoints that look like `/channels/` instead of `/channels`.

### DIFF
--- a/docs/http_api.rst
+++ b/docs/http_api.rst
@@ -118,7 +118,7 @@ Channels
 Messages
 ^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review.

rudigiesler:
Maybe we should do the same for `/channels/channel-id/messages` to be consistent?

justinvdm:
Thanks, missed that :)

justinvdm:
Done.

rudigiesler:
:+1:

PR #9, opened 2015-09-02
Send a message

rudigiesler:
Add the ability to send a message across a channel by posting to `/channels/channel-id/messages`. Adds the message to the correct rabbitmq queue for the specified channel-id.

### DIFF
--- a/junebug/amqp.py
+++ b/junebug/amqp.py
@@ -70,15 +70,15 @@ class JunebugAMQClient(AMQClient, object):
         self.factory.amqp_client_d.callback(None)
 
     @inlineCallbacks
-    def get_channel(self, channel_id=None):
-        """If channel_id is None a new channel is created"""
-        if channel_id:
-            channel = self.channels[channel_id]
-        else:
+    def get_channel(self):
+        """If channel is None a new channel is created"""
+        if not hasattr(self, 'cached_channel'):
             channel_id = self.get_new_channel_id()
             channel = yield self.channel(channel_id)
             yield channel.channel_open()
-            self.channels[channel_id] = channel
+            self.cached_channel = channel
+        else:
+            channel = self.cached_channel
         returnValue(channel)
 
     def get_new_channel_id(self):
--- a/junebug/amqp.py
+++ b/junebug/amqp.py
@@ -1,23 +1,69 @@
+from twisted.application.internet import TCPClient
+from twisted.application.service import MultiService
 from twisted.internet.defer import inlineCallbacks, returnValue, Deferred
 from twisted.internet.protocol import ReconnectingClientFactory
 from twisted.python import log
+from twisted.web import http
 from txamqp.client import TwistedDelegate
 from txamqp.content import Content
 from txamqp.protocol import AMQClient
 from vumi.utils import vumi_resource_path
 from vumi.service import get_spec
 
+from junebug.error import JunebugError
+
+
+class AmqpConnectionError(JunebugError):
+    '''Exception that is raised whenever a message is attempted to be send,
+    but no amqp connection is available to send it on'''
+    name = 'AmqpConnectionError'
+    description = 'amqp connection error'
+    code = http.INTERNAL_SERVER_ERROR
+
+
+class MessageSender(MultiService):
+    '''Keeps track of the amqp connection and can send messages. Raises an
+    exception if a message is sent when there is no amqp connection'''
+    def __init__(self, specfile, amqp_config):
+        super(MessageSender, self).__init__()
+        self.amqp_config = amqp_config
+        self.factory = AmqpFactory(
+            specfile, amqp_config, self._connected_callback,
+            self._disconnected_callback)
+
+    def startService(self):
+        super(MessageSender, self).startService()
+        self.amqp_service = TCPClient(
+            self.amqp_config['hostname'], self.amqp_config['port'],
+            self.factory)
+        self.amqp_service.setServiceParent(self)
+
+    def _connected_callback(self, client):
+        self.client = client
+
+    def _disconnected_callback(self):
+        self.client = None
+
+    def send_message(self, message, **kwargs):
+        if not hasattr(self, 'client') or self.client is None:
+            raise AmqpConnectionError(
+                'Message not sent, AMQP connection error.')
+        return self.client.publish_message(message, **kwargs)
+
 
 class AmqpFactory(ReconnectingClientFactory, object):
-    def __init__(self, specfile, amqp_config):
+    def __init__(
+            self, specfile, amqp_config, connected_callback,
+            disconnected_callback):
         '''Factory that creates JunebugAMQClients.
         specfile - string of specfile name
         amqp_config - connection details for amqp server
         '''
+        self.connected_callback, self.disconnected_callback = (
+            connected_callback, disconnected_callback)
         self.amqp_config = amqp_config
         self.spec = get_spec(vumi_resource_path(specfile))
         self.delegate = TwistedDelegate()
-        self.amqp_client_d = Deferred()
 
     def buildProtocol(self, addr):
         amqp_client = JunebugAMQClient(
@@ -35,15 +81,8 @@ class AmqpFactory(ReconnectingClientFactory, object):
     def clientConnectionLost(self, connector, reason):
         log.err("AmqpFactory client connection lost (%s)" % (
             reason.getErrorMessage(),))
-        self.amqp_client_d = Deferred()
+        self.disconnected_callback()
         super(AmqpFactory, self).clientConnectionLost(connector, reason)
-
-    @inlineCallbacks
-    def get_client(self):
-        '''Returns a deferred that fires with a connected client'''
-        if not self.amqp_client_d.called:
-            yield self.amqp_client_d
-        returnValue(self.amqp_client)
 
 
 class RoutingKeyError(Exception):
@@ -66,8 +105,7 @@ class JunebugAMQClient(AMQClient, object):
                                 self.factory.amqp_config['password'])
         # authentication was successful
         log.msg("Got an authenticated AMQP connection")
-        self.factory.amqp_client = self
-        self.factory.amqp_client_d.callback(None)
+        self.factory.connected_callback(self)
 
     @inlineCallbacks
     def get_channel(self):
--- a/junebug/amqp.py
+++ b/junebug/amqp.py
@@ -64,6 +64,7 @@ class AmqpFactory(ReconnectingClientFactory, object):
         self.amqp_config = amqp_config
         self.spec = get_spec(vumi_resource_path(specfile))
         self.delegate = TwistedDelegate()
+        super(AmqpFactory, self).__init__()
 
     def buildProtocol(self, addr):
         amqp_client = JunebugAMQClient(
--- a/junebug/amqp.py
+++ b/junebug/amqp.py
@@ -1,6 +1,6 @@
 from twisted.application.internet import TCPClient
 from twisted.application.service import MultiService
-from twisted.internet.defer import inlineCallbacks, returnValue, Deferred
+from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.protocol import ReconnectingClientFactory
 from twisted.python import log
 from twisted.web import http
--- a/junebug/amqp.py
+++ b/junebug/amqp.py
@@ -1,0 +1,115 @@
+from twisted.internet.defer import inlineCallbacks, returnValue, Deferred
+from twisted.internet.protocol import ReconnectingClientFactory
+from twisted.python import log
+from txamqp.client import TwistedDelegate
+from txamqp.content import Content
+from txamqp.protocol import AMQClient
+from vumi.utils import vumi_resource_path
+from vumi.service import get_spec
+
+
+class AmqpFactory(ReconnectingClientFactory, object):
+    def __init__(self, specfile, amqp_config):
+        '''Factory that creates JunebugAMQClients.
+        specfile - string of specfile name
+        amqp_config - connection details for amqp server
+        '''
+        self.amqp_config = amqp_config
+        self.spec = get_spec(vumi_resource_path(specfile))
+        self.delegate = TwistedDelegate()
+        self.amqp_client_d = Deferred()
+
+    def buildProtocol(self, addr):
+        amqp_client = JunebugAMQClient(
+            self.delegate, self.amqp_config['vhost'],
+            self.spec, self.amqp_config.get('heartbeat', 0))
+        amqp_client.factory = self
+        self.resetDelay()
+        return amqp_client
+
+    def clientConnectionFailed(self, connector, reason):
+        log.err("AmqpFactory connection failed (%s)" % (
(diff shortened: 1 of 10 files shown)


rudigiesler:
The large coverage decrease is mainly because I don't know if there is a sane way of testing the Amqp factory and client.

junebug/tests/helpers.py
```diff
@@ -1,13 +1,48 @@
...
+    def channel(self, id):
+        yield self.channelLock.acquire()
+        try:
+            try:
+                ch = self.channels[id]
+            except KeyError:
+                ch = FakeAMQPChannel(id, self)
+                self.channels[id] = ch
+        finally:
+            self.channelLock.release()
```
justinvdm: Could you explain why we need a lock?

junebug/amqp.py
```diff
@@ -0,0 +1,115 @@
...
+from twisted.internet.protocol import ReconnectingClientFactory
+from twisted.python import log
+from txamqp.client import TwistedDelegate
+from txamqp.content import Content
+from txamqp.protocol import AMQClient
+from vumi.utils import vumi_resource_path
+from vumi.service import get_spec
+
+
+class AmqpFactory(ReconnectingClientFactory, object):
```
justinvdm: From slack discussion, we are thinking of making a new PR later for changing the in-process reconnecting behaviour to be configurable via a config flag so that one can choose to enable or disable reconnecting (and have the process exit if reconnecting is disabled).

junebug/tests/helpers.py
rudigiesler: This is copied from vumi, so I don't really know. Basically, I needed to do the same thing that vumi did, but inherit from `JunebugAMQClient` instead, and I couldn't think of a better way of doing it than copy pasting.

justinvdm: I see. I think we should maybe try avoid duplicating stuff vumi is doing. One way could be to have our `FakeAmqpClient` mixin `JunebugAMQClient` and vumi's `FakeAMQClient`.

Alternatively, maybe if we're doing mixins, its a sign that we need to using composition instead of inheritance (so have `JunebugAMQClient` have the amqp client as an instance variable passed to it via `__init__` or something).

Thoughts?

junebug/channel.py
```diff
@@ -29,6 +30,13 @@ class InvalidChannelType(JunebugError):
     'xmpp': 'vumi.transports.xmpp.XMPPTransport',
 }
 
+allowed_message_fields = [
+    'transport_name', 'timestamp', 'in_reply_to', 'to_addr', 'from_addr',
+    'content', 'session_event', 'helper_metadata', 'message_id']
+# excluded fields: from_addr_type, group, provider, routing_metadata,
+# to_addr_type, from_addr_type, message_version, transport_metadata,
+# message_type, transport_type
```
justinvdm: If junebug should also support sending reply messages, I think we might need transport metadata. I think some transports add some metadata for inbound messages that reply messages use, like session identifiers.

Is `message_id` maybe something we want to generate ourselves.

Is `transport_name` something we need from people?

Any reason to disallow `group`?

junebug/channel.py
```diff
@@ -142,3 +152,15 @@ def status(self):
...
+    @inlineCallbacks
+    def send_message(self, amq_client, to_addr, content, **kw):
+        message = TransportUserMessage.send(
+            to_addr=to_addr, content=content, transport_name=self.id)
+        queue = '%s.outbound' % self.id
+        msg = yield amq_client.publish_message(message, routing_key=queue)
+        ret = {}
+        for key, val in msg.payload.iteritems():
+            if key in allowed_message_fields:
+                ret[key] = val
```
justinvdm: Would it maybe be worth throwing an error if disallowed fields were given? Maybe a job for jsonschema?

junebug/channel.py
rudigiesler: The `allowed_message_fields` is actually just a list of things that we return on the http request, I should probably find a better name for it. This is just a rough draft of the list, which is why I put the comment of the excluded fields, so that we don't forget about them. The idea is to reveal as little as possible on the API, to make it simpler to understand.

junebug/channel.py
rudigiesler: This has to do with what we return, not what gets sent. At the moment, only `to_addr` and `content` are sent through. I didn't want to make this PR too big with validation things, but we'll need to make sure to pass all the fields given to us in the message.

junebug/amqp.py
```diff
@@ -0,0 +1,115 @@
...
+            reason.getErrorMessage(),))
+        self.amqp_client_d = Deferred()
+        super(AmqpFactory, self).clientConnectionLost(connector, reason)
+
+    @inlineCallbacks
+    def get_client(self):
+        '''Returns a deferred that fires with a connected client'''
+        if not self.amqp_client_d.called:
+            yield self.amqp_client_d
+        returnValue(self.amqp_client)
```
justinvdm: From the slack discussion (making a comment here so I don't forget), the plan is to get the client class via a class level variable, and then override that class level variable with a dummy client class and use that as a way to test `AmqpFactory`.

junebug/amqp.py
justinvdm: Update: we are probably going to change vumi's amqp things to do what we need instead of doing the current amqp things we are doing in junebug.

junebug/channel.py
justinvdm: Ah, woops, sorry, read this wrong.

justinvdm: Ok, now that I'm not reading this wrong, do we maybe want to add validation for the message fields given to `send_message`? To disallow fields that we will be replacing anyways, or fields that people shouldn't be setting. Or maybe that's something we should leave up to the api?

justinvdm: On second thought, lets leave that validation for the api, like it is at the moment, sorry for the noise.

junebug/api.py
```diff
@@ -172,13 +179,16 @@ def delete_channel(self, request, channel_id):
             'type': 'object',
             'properties': {
                 'to': {'type': 'string'},
-                'from': {'type': 'string'},
+                'from': {'type': ['string', 'null']},
                 'reply_to': {'type': 'string'},
+                'content': {'type': 'string'},
```
justinvdm: `null` content is a valid case, I think, for cases where we start sessions.

junebug/channel.py
rudigiesler: Yes, I'm thinking of creating a function that takes a vumi message, and creates a dictionary that should be returned to the user, because some of the conversion can be a bit tricky, so I think it's best to do it manually.

justinvdm: Sounds good.

rudigiesler:
Probably ready for a proper review

junebug/api.py
```diff
@@ -172,13 +177,16 @@ def delete_channel(self, request, channel_id):
...
-                'from': {'type': 'string'},
+                'from': {'type': ['string', 'null']},
                 'reply_to': {'type': 'string'},
+                'content': {'type': ['string', 'null']},
                 'event_url': {'type': 'string'},
                 'priority': {'type': 'string'},
                 'channel_data': {'type': 'object'},
-            }
+            },
+            'required': ['from', 'content'],
```
justinvdm: Do we want to add validation for people giving fields they shouldn't be giving, for e.g. `transport_name`. People (and machines) might find it surprising that the values they gave for fields were silently overridden with our own fields.

rudigiesler: Done.

junebug/channel.py
```diff
@@ -142,3 +152,42 @@ def status(self):
...
+        ret['from'] = msg['from_addr']
+        ret['message_id'] = msg['message_id']
+        ret['channel_id'] = msg['transport_name']
+        ret['timestamp'] = msg['timestamp']
+        ret['reply_to'] = msg['in_reply_to']
+        ret['content'] = msg['content']
+        ret['session_event'] = msg['session_event']
+        ret['channel_data'] = msg['helper_metadata']
+        if not ret['channel_data']:
+            ret['channel_data'] = {}
```
justinvdm: Is this to handle cases where `helper_metadata` is `None`? If so, not sure if we have those cases.

junebug/channel.py
```diff
@@ -142,3 +152,42 @@ def status(self):
...
+        if msg.get('continue_session'):
+            ret['channel_data']['continue_session'] = msg['continue_session']
```
justinvdm: Won't we want to include this in `channel_data` even if it is falsy?

junebug/tests/helpers.py
justinvdm: If this method is identical to vumi's `FakeAMQClient`, maybe be worth having this class be a mixin of `JunebugAMQClient` and `FakeAMQClient`? Mixing in doesn't seem great, but seems better than duplicating code.

junebug/channel.py
justinvdm: Separate thing -- for what reason do we put `continue_session` in `channel_data` instead of having it be top-level like the other fields?

junebug/amqp.py
```diff
@@ -0,0 +1,153 @@
...
+    def clientConnectionFailed(self, connector, reason):
+        log.err("AmqpFactory connection failed (%s)" % (
+            reason.getErrorMessage(),))
+        super(AmqpFactory, self).clientConnectionFailed(connector, reason)
+
+    def clientConnectionLost(self, connector, reason):
+        log.err("AmqpFactory client connection lost (%s)" % (
+            reason.getErrorMessage(),))
+        self.disconnected_callback()
+        super(AmqpFactory, self).clientConnectionLost(connector, reason)
```
justinvdm: I'm wondering if it'd be worth testing that these methods do what we expect them to (i.e. that `buildProtocol` constructs the client, that the connection failed callback logs an error, and that the conection lost callback logs an error and calls a callback). I think we can assume that `ReconnectClient` will do the right thing with the methods, but still might be worth testing that the methods that we override do the right thing.

junebug/tests/test_amqp.py
```diff
@@ -0,0 +1,72 @@
+from twisted.internet.defer import inlineCallbacks
+from vumi.message import TransportUserMessage
+
+from junebug.amqp import (
+    AmqpConnectionError, AmqpFactory, JunebugAMQClient, RoutingKeyError)
+from junebug.tests.helpers import JunebugTestBase
+
+
+class TestJunebugApi(JunebugTestBase):
```
justinvdm: Should this be `TestMessageSender`?

junebug/channel.py
```diff
@@ -142,3 +152,42 @@ def status(self):
...
+
+    def _message_from_api(self, msg):
+        ret = {}
+        ret['to_addr'] = msg.get('to')
+        ret['from_addr'] = msg['from']
+        ret['content'] = msg['content']
+        ret['transport_name'] = self.id
+        channel_data = msg.get('channel_data', {})
+        if channel_data.get('continue_session'):
+            ret['continue_session'] = channel_data.pop('continue_session')
```
justinvdm: Similar comment to [here](https://github.com/praekelt/junebug/pull/9/files#r38942256).

junebug/amqp.py
```diff
@@ -0,0 +1,153 @@
...
+            disconnected_callback):
+        '''Factory that creates JunebugAMQClients.
+        specfile - string of specfile name
+        amqp_config - connection details for amqp server
+        '''
+        self.connected_callback, self.disconnected_callback = (
+            connected_callback, disconnected_callback)
+        self.amqp_config = amqp_config
+        self.spec = get_spec(vumi_resource_path(specfile))
+        self.delegate = TwistedDelegate()
```
justinvdm: Do we need to do the `super()` dance for this `__init__()`?

junebug/amqp.py
```diff
@@ -0,0 +1,153 @@
...
+class AmqpFactory(ReconnectingClientFactory, object):
+    def __init__(
+            self, specfile, amqp_config, connected_callback,
+            disconnected_callback):
+        '''Factory that creates JunebugAMQClients.
+        specfile - string of specfile name
+        amqp_config - connection details for amqp server
+        '''
+        self.connected_callback, self.disconnected_callback = (
+            connected_callback, disconnected_callback)
```
justinvdm: When do we use `connected_callback`?

justinvdm: Ah, sorry, just saw we use it via the client.

junebug/amqp.py
```diff
@@ -0,0 +1,153 @@
...
+        return self.publish(amq_message, **kwargs)
+
+    @inlineCallbacks
+    def publish(self, message, **kwargs):
+        exchange_name = kwargs.get('exchange_name') or self.exchange_name
+        routing_key = kwargs.get('routing_key') or self.routing_key
+        self.check_routing_key(routing_key)
+        channel = yield self.get_channel()
+        yield channel.basic_publish(
+            exchange=exchange_name, content=message, routing_key=routing_key)
```
justinvdm: Would it maybe be worth testing that this class's methods are doing the right thing? I think since testing via real rabbit would be more work that it is worth, stubbing `basic_publish()` and `channel()` might be enough to test that these methods are doing the right thing. Sorry for commenting on this so late, it should've been done in the first review.

justinvdm:
Made a few more comments, looks good :)

junebug/channel.py
rudigiesler: Yes you're right, I'll remove that.

junebug/channel.py
rudigiesler: Ah yes, this should be a `None` check

rudigiesler: It's what was decided when we were designing the API. All top level things are things that can apply to all transports, `channel_data` things are things that are only relevant to certain transports.

junebug/amqp.py
rudigiesler: We are testing `buildProtocol`, I'll see how difficult it is to add tests for the `client*` methods.

junebug/channel.py
justinvdm: Ah cool. I would've thought the same would be true for `session_event` then, or is it not so?

junebug/amqp.py
justinvdm: Thanks :)

rudigiesler: Turns out it's quite difficult. Tried using a `_FakeConnector`, but the `clientConnectionFailed` of the `ReconnectingClientFactory` does a `clock.callLater(...)`, which makes it quite difficult to test. Ideas on testing this?

justinvdm: I see. Hmm, maybe its not worth the trouble then, thanks for taking a look.

junebug/amqp.py
rudigiesler: Done

junebug/amqp.py
rudigiesler: Done

rudigiesler:
Okay, I think I've addressed all of the comments @justinvdm

junebug/channel.py
rudigiesler: Yes actually, `session_event` isn't in the API spec, I'll change that

justinvdm:
:+1:!

PR #16, opened 2015-09-04
Fix duplicate channel startup

rudigiesler:


justinvdm:
:+1: but maybe we want to create an issue for yielding on worker startup and teardown for `Channel.start` and `Channel.stop`?

Looks like coveralls failed because of the 0.006% decrease in coverage, maybe we should tweak it.

rudigiesler:
issue created: https://github.com/praekelt/junebug/issues/17

PR #18, opened 2015-09-09
Make send_message a class method

rudigiesler:
Currently, whenever a channel is created, it fetches its properties from redis. For some things (mainly sending messages), this isn't needed, so we should fetch properties in a more "on demand" way, caching it on the channel object.


justinvdm:
:+1:

hodgestar:
:+1: from me too.

PR #8, opened 2015-09-09
Get config from file

justinvdm:
Currently, all config options are added from the command line. The goal of this is to be able to specify a configuration file that can also set all of these options, instead of specifying them in the command line.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #20, opened 2015-09-10
add message sending application worker

rudigiesler:
We need an application worker that will take MO messages and send them to the provided URL. The plan is that a separate PR will handle the starting/stopping of the worker (when channels are started/stopped).

### DIFF
--- a/junebug/workers.py
+++ b/junebug/workers.py
@@ -1,0 +1,40 @@
+import json
+import logging
+from twisted.internet.defer import inlineCallbacks
+import treq
+from vumi.application.base import ApplicationConfig, ApplicationWorker
+from vumi.config import ConfigText
+from vumi.message import JSONMessageEncoder
+
+from junebug.channel import Channel
+
+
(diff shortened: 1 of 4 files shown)


junebug/workers.py
```diff
@@ -0,0 +1,40 @@
...
+    @inlineCallbacks
+    def consume_user_message(self, message):
+        '''Sends the vumi message as an HTTP request to the configured URL'''
+        config = yield self.get_config(message)
+        url = config.mo_message_url
+        headers = {
+            'Content-Type': 'application/json',
+        }
+        msg = json.dumps(
+            Channel._api_from_message(message), cls=JSONMessageEncoder)
```
justinvdm: Not very important, but maybe if we are using this method as a public thing it shouldn't have a `_` prefix?

justinvdm:
One minor thing, looks good :)

rudigiesler:
Changes made. Ready for review again

justinvdm:
:+1:

PR #21, opened 2015-09-11
Create, start and stop application workers with their transport workers

justinvdm:
#20 adds an application worker for `POST`ing mo messages to a url. We plan to have one application worker instance per transport instance, and create, start and stop each application worker with its corresponding transport worker.

### DIFF
--- a/junebug/tests/test_channel.py
+++ b/junebug/tests/test_channel.py
@@ -61,8 +61,11 @@ class TestChannel(JunebugTestBase):
 
         self.assertTrue(isinstance(worker, MessageForwardingWorker))
         self.assertEqual(self.service.namedServices[id], worker)
-        self.assertEqual(worker.config['transport_name'], channel.id)
-        self.assertEqual(worker.config['mo_message_url'], 'http://foo.org')
+
+        self.assertEqual(worker.config, {
+            'transport_name': channel.id,
+            'mo_message_url': 'http://foo.org'
+        })
 
     @inlineCallbacks
     def test_create_channel_invalid_type(self):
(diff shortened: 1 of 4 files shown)


justinvdm:
Ready for review, though I still need to figure out why the application worker service isn't being run.

junebug/tests/test_channel.py
```diff
@@ -72,6 +91,43 @@ def test_update_channel(self):
...
+        self.assertEqual(self.service.namedServices[channel.id], worker1)
+        yield channel.update({'foo': 'bar'})
+        self.assertEqual(self.service.namedServices[channel.id], worker1)
+
+        config = self.create_channel_config()
+        config['config']['foo'] = ['bar']
+        yield channel.update(config)
+
+        worker2 = channel.transport_worker
+        self.assertEqual(self.service.namedServices[channel.id], worker2)
```
rudigiesler: Should we maybe assert that `worker1` is no longer in the list of services?

junebug/tests/test_channel.py
```diff
@@ -72,6 +91,43 @@ def test_update_channel(self):
...
+        yield channel.update({'foo': 'bar'})
+        self.assertEqual(self.service.namedServices[id], worker1)
+
+        config = self.create_channel_config()
+        config['config']['foo'] = ['bar']
+        yield channel.update(config)
+
+        worker2 = channel.application_worker
+        self.assertEqual(self.service.namedServices[id], worker2)
+        self.assertNotEqual(worker1, worker2)
```
rudigiesler: Shouldn't we also be restarting the application if the `mo_url` attribute changes? Not only if the `config` (which is the transport config) changes.

rudigiesler:
Just those two comments from me, looks good otherwise

junebug/tests/test_channel.py
justinvdm: Good point, will change.

justinvdm:
Ready for re-review.

rudigiesler:
:+1:

PR #23, opened 2015-09-11
Fix mo_url in message forwarding worker

rudigiesler:
Currently, getting the `mo_url` from the config results in unicode. We must convert this to bytes before trying to make an http request.

### DIFF
--- a/junebug/tests/test_workers.py
+++ b/junebug/tests/test_workers.py
@@ -49,7 +49,7 @@ class TestMessageForwardingWorker(JunebugTestBase):
 
         app_config = {
             'transport_name': 'testtransport',
-            'mo_message_url': self.url,
+            'mo_message_url': unicode(self.url),
         }
         self.worker = yield self.get_worker(app_config)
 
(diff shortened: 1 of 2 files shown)


junebug/tests/test_workers.py
```diff
@@ -49,7 +49,7 @@ def setUp(self):
 
         app_config = {
             'transport_name': 'testtransport',
-            'mo_message_url': self.url,
+            'mo_message_url': unicode(self.url),
```
hodgestar: We should probably use `.decode("utf-8")` instead of `unicode(...)` which uses the system encoding (which could be almost anything but is probably ASCII).

junebug/workers.py
```diff
@@ -27,7 +27,7 @@ class MessageForwardingWorker(ApplicationWorker):
     def consume_user_message(self, message):
         '''Sends the vumi message as an HTTP request to the configured URL'''
         config = yield self.get_config(message)
-        url = config.mo_message_url
+        url = str(config.mo_message_url)
```
hodgestar: Here we should use `.encode("utf-8")`.

rudigiesler:
@hodgestar Ready for rereview.

hodgestar:
:+1:

PR #22, opened 2015-09-14
Don't patch transport worker creation in api tests

justinvdm:
It would be cool if we could let `Channel` do its usual transport creation.


justinvdm:
Ready for review.

Part of this PR ended up being making tests more consistent with [this relevant test change from #21](https://github.com/praekelt/junebug/commit/102f9be4aea3a9005d594ad1faec38e7082a2cab#comments). Hopefully the cleanup was relevant enough.

rudigiesler:
:+1:

PR #24, opened 2015-09-14
Remove channel from channel list when deleting

rudigiesler:
Currently, we delete the channel data from redis, but we don't remove the channel id from the channel list.


justinvdm:
:+1: when travis is happy.

PR #25, opened 2015-09-15
Create inbound message store

rudigiesler:
For inbound messages, we need to store the message so that we can construct a reply_to message. The message will be stored in redis with an expiry.

### DIFF
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -31,10 +31,9 @@ class BaseStore(object):
     def load_property(self, prop):
         self.properties[prop] = yield self.redis.hget(self.id, prop)
 
-    @inlineCallbacks
     def set_property(self, prop, value):
-        yield self.redis.hset(self.id, prop, value)
         self.properties[prop] = value
+        return succeed(None)
 
     def get_property(self, prop):
         return succeed(self.properties.get(prop))
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -1,0 +1,38 @@
+from twisted.internet.defer import inlineCallbacks, succeed, returnValue
+
+
+class BaseStore(object):
+    '''Base class for store objects. Stores data in redis in a hash.
+    redis: redis manager
+    ttl: expiry for the key
+    id: key to store the hash under
+    properties: the keys and values of the hash
+    '''
+
+    def __init__(self, id, properties, redis, ttl):
+        self.id, self.properties, self.redis, self.ttl = (
+            id, properties, redis, ttl)
+
+    @inlineCallbacks
+    def save_all(self):
+        yield self.redis.hmset(self.id, self.properties)
+        yield self.redis.expire(self.id, self.ttl)
+
+    @inlineCallbacks
+    def load_all(self):
+        self.properties = (yield self.redis.hgetall(self.id)) or {}
+
+    @inlineCallbacks
+    def set_property(self, prop, value):
+        yield self.redis.hset(self.id, prop, value)
+        self.properties[prop] = value
+
+    def get_property(self, prop):
+        return succeed(self.properties.get(prop))
+
+    @classmethod
+    @inlineCallbacks
+    def from_id(cls, id, redis, ttl):
+        store = cls(id, {}, redis, ttl)
+        yield store.load_all()
+        returnValue(store)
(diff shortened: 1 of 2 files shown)


rudigiesler:
Ready for review.

junebug/stores.py
```diff
@@ -0,0 +1,64 @@
...
+    '''Raised when a message cannot be found.'''
+    name = 'MessageNotFound'
+    description = 'message not found'
+    code = http.NOT_FOUND
+
+
+class BaseStore(object):
+    '''Base class for store objects. Stores data in redis in a hash.
+    redis: redis manager
+    ttl: expiry for the key
```
justinvdm: Does this docstring need to be updated to suggest that the ttl is for all keys managed by this store?

justinvdm:
Wondering if it would make sense to add a prefix for keys added by the store? So we could add a prefix like `inbound.` for the inbound store. Might help for working with/inspecting/scanning a store's keys later on.

rudigiesler:
My idea is that since you pass through a redis manager, that could also be a submanager, so you could control the prefix through that.

junebug/stores.py
```diff
@@ -0,0 +1,64 @@
...
+            message.get('message_id'), 'message', message.to_json())
+
+    @inlineCallbacks
+    def load_vumi_message(self, message_id):
+        '''Retrieves the stored vumi message, given its unique id'''
+        msg_json = yield self.load_property(message_id, 'message')
+        if msg_json is None:
+            raise MessageNotFound(
+                'Cannot find message with id %s' % (message_id,))
+        returnValue(TransportUserMessage.from_json(msg_json))
```
justinvdm: `load_property` returns `None` for non-existent properties, but `load_vumi_message` raises an error. I think we should maybe make the behavior of load methods more consistent (so they either all return `None` when a thing doesn't exist, or they all throw an error).

justinvdm:
> My idea is that since you pass through a redis manager, that could also be a submanager, so you could control the prefix through that.

Ah, +1

justinvdm:
Left some minor comments, looks good :)

junebug/stores.py
rudigiesler: Which one would be your preference?

justinvdm: I don't have much of a preference, happy to go with what you think makes more sense :) Maybe returning `None` makes sense, since it matches up with the way we interact with redis?

rudigiesler:
Comments addressed. Ready for another review.

junebug/stores.py
```diff
@@ -0,0 +1,53 @@
+from twisted.internet.defer import inlineCallbacks, returnValue
+from vumi.message import TransportUserMessage
+
+
+class BaseStore(object):
+    '''Base class for store objects. Stores data in redis in a hash.
+    redis: redis manager
+    ttl: expiry for keys in the store
```
hodgestar: Should we update this docstring to use the usual Sphinx formatting? That way if we ever pull it into autodoc it'll just work (plus it'll be more consistent with our docs elsewhere).

junebug/stores.py
```diff
@@ -0,0 +1,53 @@
...
+    '''
+
+    def __init__(self, redis, ttl):
+        self.redis, self.ttl = redis, ttl
```
hodgestar: Bikeshed alert: I prefer these on separate lines (feel free to ignore).

junebug/stores.py
rudigiesler: Does sphinx require this to be the class docstring or the `__init__` method docstring?

hodgestar: The standard seems to be to put it on the class and to leave the docstring
for __init__ empty.

rudigiesler:
Docstring updated to use sphinx formatting.

junebug/stores.py
```diff
@@ -0,0 +1,59 @@
+from twisted.internet.defer import inlineCallbacks, returnValue
+from vumi.message import TransportUserMessage
+
+
+class BaseStore(object):
+    '''
+    .. class:: junebug.stores.BaseStore(redis, ttl)
```
hodgestar: I think autodoc doesn't need the `.. class:: ...` -- it can figure that out from the fact that this is a class.

hodgestar:
Left one additional docstring comment, otherwise looks good to me.

justinvdm:
:+1:

hodgestar:
:+1:

PR #26, opened 2015-09-16
Store incoming messages

rudigiesler:


junebug/channel.py
```diff
@@ -193,9 +193,17 @@ def _transport_config(self):
 
     @property
     def _application_config(self):
+        redis_manager = self.redis._config
+        # For tests, FAKE_REDIS cannot be an object, else deepcopy doesn't
+        # work.
+        if 'FAKE_REDIS' in redis_manager:
+            redis_manager['FAKE_REDIS'] = 'yes'
```
rudigiesler: I'm looking for a better way to do this, suggestions welcome.

junebug/config.py
```diff
@@ -34,3 +34,7 @@ class JunebugConfig(Config):
             'username': 'guest',
             'password': 'guest'
         })
+
+    ttl = ConfigInt(
+        "Maximum time allowed to reply to messages",
+        default=60)
```
justinvdm: Since we are going to have a ttl for storing outbound messages, maybe we should name this to `inbound_ttl`?

rudigiesler: I was wondering about that. Should it be the same ttl, or should we have separate ttls?

junebug/workers.py
```diff
@@ -24,8 +30,22 @@ class MessageForwardingWorker(ApplicationWorker):
     CONFIG_CLASS = MessageForwardingConfig
 
     @inlineCallbacks
+    def setup_application(self):
+        self.redis = yield TxRedisManager.from_config(
+            self.config['redis_manager'])
+        message_redis = self.redis.sub_manager(
+            '%s:incoming_messages' % self.config['transport_name'])
```
justinvdm: Should we maybe use `inbound` or `inbound_messages` for consistency?

justinvdm: Separate thing -- we need to use the same prefix creating the inbound store on `JunebugApi` for loading messages, so I think we need a place to put the prefix that both the api and this worker can read from. Maybe we should make the prefix be a config field? That way the application worker's can be configured to use the prefix we use in `JunebugApi`.

rudigiesler: Okay, I'll put it in the config field of the worker.

junebug/channel.py
justinvdm: Would we maybe be able to get away with not needing this because of [this](https://github.com/praekelt/vumi/blob/develop/vumi/persist/redis_base.py#L70-L72)? Or did you come across an issue before adding these lines?

justinvdm:
Left a few comments, looks good :)

junebug/config.py
justinvdm: Hmm. I'd imagine it would have the same ttl value, but maybe for some reason we want to set one of the ttls explicitly?

junebug/channel.py
```diff
@@ -43,13 +43,13 @@ class Channel(object):
     APPLICATION_CLS_NAME = 'junebug.workers.MessageForwardingWorker'
 
     def __init__(
-            self, redis_manager, amqp_config, properties, id=None):
+            self, redis_manager, amqp_config, properties, ttl, id=None):
```
justinvdm: This `__init__` is also getting pretty long, wondering if we should also pass it the/a config too?

junebug/channel.py
rudigiesler: Yes, came across this issue before adding those lines. We get the `deepcopy` from confmodel, and if `FAKE_REDIS` is an object instead of a string, then it fails.

rudigiesler:
So I've change things around so that now the entire config is passed down, rather than bits of it. This should make it a lot easier to add something to the config, and have it available everywhere. Ready for another review.

junebug/channel.py
justinvdm: Ah, I see. Hmm, wondering if it would okay to just use `self.config.redis` here for `redis_manager`. I might be missing a reason why that wouldn't work though.

junebug/command_line.py
```diff
@@ -61,6 +61,14 @@ def parse_arguments(args):
     parser.add_argument(
         '--amqp-password', '-amqppass', dest='amqp_pass', type=str,
         help='The password to use for the amqp auth. Defaults to "guest"')
+    parser.add_argument(
+        '--message-ttl', '-ttl', dest='ttl', type=int,
+        help='The maximum time allowed to reply to a message. Defaults to '
+        '"60" seconds')
```
justinvdm: If we're going to be using the same ttl for both the inbound and outbound store, maybe the description should rather say something about the maximum time we store messages for. If we're using different ttls for each, maybe the option should be named to suggest it is for stored inbound messages. Similarly for the config field.

junebug/workers.py
```diff
@@ -24,8 +33,24 @@ class MessageForwardingWorker(ApplicationWorker):
     CONFIG_CLASS = MessageForwardingConfig
 
     @inlineCallbacks
+    def setup_application(self):
+        self.redis = yield TxRedisManager.from_config(
+            self.config['redis_manager'])
+        message_redis = self.redis.sub_manager(
+            '%s:%s' % (
+                self.config['transport_name'],
+                self.config['inbound_message_prefix']))
```
justinvdm: I think we usually include the `:` in the prefix itself. Would also be slightly DRYer, I guess, in case we changed the delimiter in one place but not the other(s).

justinvdm:
Left a few more comments. The pass-config-to-all-the-things change looks good :)

junebug/workers.py
rudigiesler: I'm not sure that I understand this comment?

junebug/channel.py
rudigiesler: I think that will work. The reason it didn't work before is that we weren't passing the whole config to the channel, so we needed to do things like this. I'll try and see if I can correct this.

junebug/workers.py
justinvdm: Err, sorry, read these lines wrong, ignore my comment.

justinvdm:
:+1:

junebug/command_line.py
hodgestar: I would vote for `inbound-message-ttl` and `outbound-message-ttl` being separate options (given our experience with SMPP ttls).

junebug/command_line.py
```diff
@@ -61,6 +61,14 @@ def parse_arguments(args):
...
         '--amqp-password', '-amqppass', dest='amqp_pass', type=str,
         help='The password to use for the amqp auth. Defaults to "guest"')
+    parser.add_argument(
+        '--inbound-message-ttl', '-ittl', dest='inbound_message_ttl', type=int,
+        help='The maximum time allowed to reply to a message. Defaults to '
+        '"60" seconds')
+    parser.add_argument(
+        '--inbound-message-prefix', '-imp', dest='inbound_message_prefix',
+        type=str, help='The default redis key prefix to use for storing '
+        'inbound messages. Defaults to "inbound_messages"')
```
hodgestar: Do we really want the inbound message prefix to be configurable? If the top-level prefix is already configurable via the redis config, it seems better for this part to just be constant?

justinvdm:
:+1: when travis is happy

PR #28, opened 2015-09-18
have the store take the channel id

rudigiesler:
Have the message store take in the channel id for its methods, and have it create its own sub managers.


rudigiesler:
Ready for review

justinvdm:
:+1:

PR #29, opened 2015-09-21
Create outbound message store

rudigiesler:
Create an outbound message store that stores the link between the unique message id and the event URL.

### DIFF
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -1,4 +1,4 @@
-from twisted.internet.defer import inlineCallbacks, returnValue, succeed
+from twisted.internet.defer import inlineCallbacks, returnValue
 from vumi.message import TransportUserMessage
 
 
@@ -76,13 +76,10 @@ class OutboundMessageStore(BaseStore):
         return super(OutboundMessageStore, self).get_key(
             channel_id, 'outbound_messages', message_id)
 
-    def store_vumi_message(self, channel_id, api_request, message):
+    def store_event_url(self, channel_id, event_url, message_id):
         '''Stores the event_url'''
-        if api_request.get('event_url') is not None:
-            key = self.get_key(channel_id, message.get('message_id'))
-            return self.store_property(
-                key, 'event_url', api_request['event_url'])
-        return succeed(None)
+        key = self.get_key(channel_id, message_id)
+        return self.store_property(key, 'event_url', event_url)
 
     def load_event_url(self, channel_id, message_id):
         '''Retrieves a stored event url, given the channel and message ids'''
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -11,7 +11,6 @@ class BaseStore(object):
     :param ttl: Expiry time for keys in the store
     :type ttl: integer
     '''
-    KEYSPACE = 'basestore'
 
     def __init__(self, redis, ttl):
         self.redis = redis
@@ -23,10 +22,9 @@ class BaseStore(object):
         yield self.redis.expire(id, self.ttl)
         returnValue(val)
 
-    def get_key(self, channel_id, message_id):
-        '''Gets the key where a message would be stored, using the channel and
-        message ids'''
-        return '%s:%s:%s' % (channel_id, self.KEYSPACE, message_id)
+    def get_key(self, *args):
+        '''Returns a key given strings'''
+        return ':'.join(args)
 
     def store_all(self, id, properties):
         '''Stores all of the keys and values given in the dict `properties` as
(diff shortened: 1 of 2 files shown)


rudigiesler:
Ready for review.

junebug/stores.py
```diff
@@ -1,15 +1,16 @@
-from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.defer import inlineCallbacks, returnValue, succeed
 from vumi.message import TransportUserMessage
 
 
 class BaseStore(object):
+    KEYSPACE = 'basestore'
```
hodgestar: This should go below the docstring, otherwise the docstring is not a docstring. :)

junebug/stores.py
```diff
@@ -21,6 +22,11 @@ def _redis_op(self, func, id, *args, **kwargs):
         yield self.redis.expire(id, self.ttl)
         returnValue(val)
 
+    def get_key(self, channel_id, message_id):
+        '''Gets the key where a message would be stored, using the channel and
+        message ids'''
+        return '%s:%s:%s' % (channel_id, self.KEYSPACE, message_id)
```
hodgestar: I'm wondering if `get_key` should just be implemented by sub-classes? The key returned seems use-case specific.

junebug/stores.py
```diff
@@ -41,21 +47,37 @@ def load_property(self, id, key):
...
         if msg_json is None:
             returnValue(None)
         returnValue(TransportUserMessage.from_json(msg_json))
+
+
+class OutboundMessageStore(BaseStore):
+    KEYSPACE = 'outbound_messages'
+    '''Stores the event url, in order to look it up when deciding where events
+    should go'''
+    def store_vumi_message(self, channel_id, api_request, message):
```
hodgestar: Should be named `store_event_url`?

hodgestar: Would be happier if this just took just the `event_url` rather than the `api_request`.

junebug/stores.py
hodgestar: I'm kind of thinking we want to prefix this key with `channels:` in case we need to store things that are not related to a particular channel at some point.

hodgestar:
Left some comments but looks good.

junebug/stores.py
rudigiesler: I did it this way to keep consistency. Currently we use `<channel-id>:properties` to store the channel properties. Should we be changing that to `channels:<channel-id>:properties`?

junebug/stores.py
rudigiesler: The idea I had doing it this way is that we might want to store other things in the future, and then we wouldn't need to change all the places that this is called, because all the information is here already (say if we decided we also wanted to store the content). Then we would just change the call to a `store_all` with a dictionary, and we keep it to only one redis call.

We could change it to just store the event url, if you think that's a better solution for now.

junebug/stores.py
rudigiesler: The reason I changed to putting it by the base class is that I would be repeating essentially the same thing in all the subclasses, and I wanted to reduce that repetition. Maybe the base class should have a method that takes in n strings, and returns a key, which the subclasses can use in their own key making functions.

junebug/stores.py
```diff
@@ -43,19 +50,37 @@ def load_property(self, id, key):
...
+    should go'''
+    KEYSPACE = 'outbound_messages'
+
+    def store_vumi_message(self, channel_id, api_request, message):
+        '''Stores the event_url'''
+        if api_request.get('event_url') is not None:
+            key = self.get_key(channel_id, message.get('message_id'))
+            return self.store_property(
+                key, 'event_url', api_request['event_url'])
+        return succeed(None)
```
justinvdm: Wondering whether stores should know about api requests and how they look. Would it maybe be better to pass in `event_url` directly?

junebug/stores.py
```diff
@@ -43,19 +50,37 @@ def load_property(self, id, key):
...
         returnValue(TransportUserMessage.from_json(msg_json))
+
+
+class OutboundMessageStore(BaseStore):
+    '''Stores the event url, in order to look it up when deciding where events
+    should go'''
+    KEYSPACE = 'outbound_messages'
+
+    def store_vumi_message(self, channel_id, api_request, message):
+        '''Stores the event_url'''
```
justinvdm: If the plan is for this method to only handle storing event urls, maybe the name should suggest that it stores event urls associated to a message?

justinvdm: Ah, sorry, just saw [this discussion](https://github.com/praekelt/junebug/pull/29#discussion_r39950445)

justinvdm:
Left minor comment, looks good :)

junebug/stores.py
hodgestar: Sounds good.

rudigiesler:
Ready for another review

justinvdm:
:+1:

rudigiesler:
Last commit changed the parameter order. Ready for review again.

hodgestar:
:+1:

PR #27, opened 2015-09-21
Supporting sending replies

justinvdm:
This and #26 will make replying possible.

### DIFF
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -9,7 +9,13 @@ from vumi.service import WorkerCreator
 from vumi.servicemaker import VumiOptions
 
 from junebug.error import JunebugError
-from junebug.stores import MessageNotFound
+
+
+class MessageNotFound(JunebugError):
+    '''Raised when a message is not found.'''
+    name = 'MessageNotFound'
+    description = 'message not found'
+    code = http.BAD_REQUEST
 
 
 class ChannelNotFound(JunebugError):
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -39,6 +39,7 @@ allowed_message_fields = [
 
 
 class Channel(object):
+    OUTBOUND_QUEUE = '%s.outbound'
     APPLICATION_ID = 'application:%s'
     APPLICATION_CLS_NAME = 'junebug.workers.MessageForwardingWorker'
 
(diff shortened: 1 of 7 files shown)


justinvdm:
Ready for review.

junebug/stores.py
```diff
@@ -1,6 +1,17 @@
...
 from vumi.message import TransportUserMessage
 
+from junebug.error import JunebugError
+
+
+class MessageNotFound(JunebugError):
+    '''Raised when a message is not found.'''
+    name = 'MessageNotFound'
+    description = 'message not found'
+    code = http.NOT_FOUND
```
justinvdm: Not sure if this is the right code to use. I think 404 responses are usually for telling the client that the requested entity was not found, not that something needed to fulfill the request was not found.

rudigiesler: So the action is "reply to this message" and we're saying "that message doesn't exist". So I would say it's a 400 Bad Request, but I think it would be good to have a discussion on this.

junebug/channel.py
```diff
@@ -144,13 +146,31 @@ def get_all(cls, redis):
...
-        '''Sends a message. Takes a junebug.amqp.MessageSender instance to
+    def send_message(cls, id, sender, msg):
+        '''Sends a message. Takes a :class:`junebug.amqp.MessageSender` instance to
         send a message.'''
-        message = TransportUserMessage.send(
-            **cls.message_from_api(id, msg))
-        queue = '%s.outbound' % id
-        msg = yield message_sender.send_message(message, routing_key=queue)
+        msg = cls.message_from_api(id, msg)
+        msg.setdefault('to_addr', None)
```
rudigiesler: Is this required? In `message_from_api` we set `to_addr` to `msg.get('to')` which will return `None` if `to` doesn't exist.

junebug/stores.py
rudigiesler: Is there a reason why this is in `stores`, but not used in `stores` and used in `channel`?

rudigiesler:
Just those two comments. Looking good to me otherwise.

junebug/stores.py
justinvdm: +1 on 400.

Ya, wasn't too sure where to put the error. I put it in `stores` in case something else irrelevant to `channel` was using `stores` needed to raise such an error. Happy to move it into `channel` though.

Might be a sign that we need to raise an error in the stores methods (counter to what I suggested in an earlier PR review), but not too sure, still might seem weird to have some of our methods throwing errors and some returning `None`.

rudigiesler: It might also be a bit strange having the store raise a 400 error if it can't find something.

justinvdm: Looking at the [spec](https://tools.ietf.org/html/rfc7231#section-6.5.1), `400` looks appropriate:
```
The 400 (Bad Request) status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing)
```

Rather than

```
The 404 (Not Found) status code indicates that the origin server did not find a current representation for the target resource or is not willing to disclose that one exists.
```

Will go with `404` and move the error to `channel` for now.

justinvdm:
Ready for re-review.

rudigiesler:
:+1:

PR #30, opened 2015-09-22
Store event urls

rudigiesler:
For outbound (MT) messages, we should store the mapping between the event url and the message id, using the OutboundMessageStore, so that we know where to send events to.

### DIFF
--- a/junebug/config.py
+++ b/junebug/config.py
@@ -41,4 +41,4 @@ class JunebugConfig(Config):
 
     outbound_message_ttl = ConfigInt(
         "Maximum time allowed for events to arrive for messages",
-        default=60*60*24*2)
+        default=60 * 60 * 24 * 2)
--- a/junebug/config.py
+++ b/junebug/config.py
@@ -36,9 +36,9 @@ class JunebugConfig(Config):
         })
 
     inbound_message_ttl = ConfigInt(
-        "Maximum time allowed to reply to messages. Defaults to 60s",
+        "Maximum time allowed to reply to messages",
         default=60)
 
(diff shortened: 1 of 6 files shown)


rudigiesler:
Ready for review.

junebug/config.py
```diff
@@ -38,3 +38,7 @@ class JunebugConfig(Config):
     inbound_message_ttl = ConfigInt(
         "Maximum time allowed to reply to messages",
         default=60)
+
+    outbound_message_ttl = ConfigInt(
+        "Maximum time allowed for events to arrive for messages",
```
justinvdm: Would it be worth mentioning the time unit in the description (similarly for the inbound ttl, sorry for not noticing earlier)?

hodgestar:
:+1:

junebug/config.py
rudigiesler: Does the `default` value not come up on the confmodel description of the field?

justinvdm: Err, not sure I'm understanding. I think my question was a little vague, was asking if it would be worth mentioning in the description that the ttls are in seconds.

rudigiesler: Ah I see it doesn't. Will add.

justinvdm: Thanks :)

justinvdm:
:+1: after we've mentioned that the ttls are in seconds.

rudigiesler:
Change made, ready for review.

justinvdm:
:+1:

rudigiesler:
(hopefully) final review

PR #32, opened 2015-09-23
Set and document ttls

rudigiesler:
For storing event URLs, the default should be 2 days, for replies, the default should be 10 minutes. These defaults and their implications should be documented, and the defaults should be set correctly.

### DIFF
--- a/docs/http_api.rst
+++ b/docs/http_api.rst
@@ -132,9 +132,13 @@ Messages
        the uuid of the message being replied to if this is a response to a
        previous message. Important for session-based transports like USSD.
        Optional. Only one of ``to`` or ``reply_to`` may be specified.
+       Defaults settings allow 10 minutes to reply to a message, after which
+       an error will be returned.
    :param str event_url:
        URL to call for status events (e.g. acknowledgements and
-       delivery reports) related to this message.
+       delivery reports) related to this message. Default settings allow for
+       2 days for events to arrive, after which they will no longer be
+       forwarded.
    :param int priority:
        Delivery priority from 1 to 5. Higher priority messages are delivered first.
        If omitted, priority is 1.
@@ -218,6 +222,8 @@ format:
 
 Events are posted to the message’s ``event_url`` after the message is
 submitted to the provider, and when delivery reports are received.
+Default settings allow events to arrive for up to 2 days; any further events
+will not be forwarded.
 
 **Request example**:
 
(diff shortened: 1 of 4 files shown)


rudigiesler:
Originally I also wanted to update the optional arguments for running `jb`, but it turns out this is complex due do sphinx not wanting to have multiple character options with a single `-`, i.e. sphinx breaks when you have `-redish`. This will be done in another PR when we've found a solution.

docs/http_api.rst
```diff
@@ -132,9 +132,13 @@ Messages
        the uuid of the message being replied to if this is a response to a
        previous message. Important for session-based transports like USSD.
        Optional. Only one of ``to`` or ``reply_to`` may be specified.
+       Defaults settings allow 10 minutes to reply to a message, after which
```
hodgestar: s/Defaults/Default/ or maybe `The default ...`?

junebug/command_line.py
```diff
@@ -64,7 +64,7 @@ def parse_arguments(args):
     parser.add_argument(
         '--inbound-message-ttl', '-ittl', dest='inbound_message_ttl', type=int,
         help='The maximum time allowed to reply to a message. Defaults to '
-        '"60" seconds')
+        '60*10 seconds or 10 minutes')
```
hodgestar: I think we need to reword this to say something like `The maximum time allowed to reply to a message (in seconds). Defaults to 600 seconds (10 minutes)`, otherwise it becomes unclear what the units are.

docs/http_api.rst
```diff
@@ -132,9 +132,13 @@ Messages
        the uuid of the message being replied to if this is a response to a
        previous message. Important for session-based transports like USSD.
        Optional. Only one of ``to`` or ``reply_to`` may be specified.
+       The default settings allow 10 minutes to reply to a message, after which
+       an error will be returned.
    :param str event_url:
        URL to call for status events (e.g. acknowledgements and
-       delivery reports) related to this message.
+       delivery reports) related to this message. Default settings allow for
```
hodgestar: Do we want to change this to `The default settings ...` too? And the one further down?

hodgestar:
Very minor comments, otherwise looks good.

rudigiesler:
Comments should be addressed.

hodgestar:
:+1:

PR #33, opened 2015-09-23
Fix message forwarding worker to ignore events without event urls

justinvdm:
At the moment, giving event urls is optional when sending a message, yet we log an error in the message forwarding worker if no event url is found.


justinvdm:
Ready for review.

rudigiesler:
:+1:

PR #34, opened 2015-09-23
Store outbound message event urls before publish outbound messages

justinvdm:
At the moment we publish first and then store, which can lead to race conditions where have not yet stored the message when handling events for the outbound message.

### DIFF
--- a/junebug/tests/test_api.py
+++ b/junebug/tests/test_api.py
@@ -317,6 +317,7 @@ class TestJunebugApi(JunebugTestBase):
             'test-channel', message['message_id'])
         self.assertEqual(event_url, None)
 
+    timeout = 1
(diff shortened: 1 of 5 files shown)


justinvdm:
Ready for review.

junebug/tests/test_api.py
```diff
@@ -317,6 +317,7 @@ def test_send_message(self):
             'test-channel', message['message_id'])
         self.assertEqual(event_url, None)
 
+    timeout = 1
```
hodgestar: Debug timeout?

justinvdm: Woops, thanks

hodgestar:
:+1:

PR #36, opened 2015-09-28
Character limits

rudigiesler:
The character limits specified on the channel config should be checked before a message is sent, and an http error should be returned if the character limit is exceeded.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -220,11 +220,10 @@ class JunebugApi(object):
 
         if 'to' in body:
             msg = yield channel.send_message(
-                channel_id, self.message_sender, self.outbounds, body)
+                self.message_sender, self.outbounds, body)
         else:
             msg = yield channel.send_reply_message(
-                channel_id, self.message_sender, self.outbounds,
-                self.inbounds, body)
+                self.message_sender, self.outbounds, self.inbounds, body)
 
         returnValue(response(request, 'message sent', msg))
 
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -6,7 +6,7 @@ from twisted.web import http
 from vumi.persist.txredis_manager import TxRedisManager
 
 from junebug.amqp import MessageSender
-from junebug.channel import Channel, ChannelNotFound
+from junebug.channel import Channel
 from junebug.error import JunebugError
 from junebug.utils import json_body, response
 from junebug.validate import body_schema, validate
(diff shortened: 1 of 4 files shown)


rudigiesler:
Ready for review.

hodgestar:
:+1:

junebug/api.py
```diff
@@ -215,10 +215,17 @@ def send_message(self, request, body, channel_id):
...
+            self.redis, self.config, channel_id, self.service)
+
+        if (channel.character_limit is not None
+                and len(body['content']) > channel.character_limit):
+            raise ApiUsageError(
+                'Message content %r is of length %d, which is greater than the'
+                ' character limit of %d' % (
+                    body['content'], len(body['content']),
+                    channel.character_limit)
+                )
```
justinvdm: I'm wondering if this logic should live in the `Channel.send_*` messages. At the moment `Channel` seems to contain the logic for how to use its config properties.

justinvdm:
Looks good, left a minor comment.

junebug/api.py
rudigiesler: The problem with using `Channel.send_*` is that they are class methods, so we would have to send through a redis manager in the `send_*` methods.

The reason they were class methods in the first place is so that we don't need to get the channel properties from redis, but since we need to do it now anyway, maybe we should look at making them normal methods?

justinvdm: Good point, +1 on making them normal methods.

rudigiesler:
Ready for review again.

junebug/channel.py
```diff
@@ -151,19 +162,17 @@ def get_all(cls, redis):
...
     @inlineCallbacks
-    def send_message(cls, id, sender, outbounds, msg):
+    def send_message(self, id, sender, outbounds, msg):
         '''Sends a message.'''
         event_url = msg.get('event_url')
         msg = message_from_api(id, msg)
         msg = TransportUserMessage.send(**msg)
-        msg = yield cls._send_message(id, sender, outbounds, event_url, msg)
+        msg = yield self._send_message(id, sender, outbounds, event_url, msg)
         returnValue(api_from_message(msg))
```
justinvdm: I think we don't need `id` anymore know that we have that as an attribute on the instance.

rudigiesler:
`id` argument removed for the message sending methods. Ready for review again.

justinvdm:
:+1:

PR #37, opened 2015-09-29
Generic channel support

rudigiesler:
The plan is to have a default list of mappings between friendly channel types, and python classes. This list can either be appended to, or replaced, via config.


rudigiesler:
Ready for review.

justinvdm:
:+1:

PR #39, opened 2015-10-05
Store events

rudigiesler:
Now that the message store allows storing of events, we must store them when we get them in the worker.


rudigiesler:
Ready for review.

justinvdm:
:+1:

PR #40, opened 2015-10-06
message status endpoint

rudigiesler:
Now that the events for messages are being stored, we can implement the message status endpoint.


rudigiesler:
Ready for review.

justinvdm:
:+1:

PR #41, opened 2015-10-13
Health event documentation updates

justinvdm:


justinvdm:
Ready for review. Struggled a bit to document status events, up for suggestions on better ways of documenting them.

docs/http_api.rst
```diff
@@ -252,3 +272,62 @@ Sent later when (or if) delivery reports are received:
...
+      purposes. For example, if the event was related to a component
+      establishing a connection, the host and port are possible fields.
+
+
+**Request Example**:
+
+.. sourcecode:: json
+
+   {
+      component: 'smpp',
```
hodgestar: I think we need to include some indication of which channel the event is from here (perhaps the worker name and/or the channel UUID). Different channels might direct their events to different URLs, but they also might not.

justinvdm: Good point, will add the channel id to be consistent with the events we send to `event_url`.

hodgestar:
Left one comment but generally looks good.

Re struggling to document things -- I think it'll be most productive to polish the documentation as a whole at the end of the current work once we all the material in place, and we do have some time allocated for that in the last of the sprints.

justinvdm:
> Re struggling to document things -- I think it'll be most productive to polish the documentation as a whole at the end of the current work once we all the material in place, and we do have some time allocated for that in the last of the sprints.
+1

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #42, opened 2015-10-16
Change documented status events to only have a single reason

justinvdm:
At the moment, status events can have multiple reasons. Multiple reasons makes sense for the request-reply pattern, since a request might fail for multiple reasons (for e.g. multiple validation failures), but for the case of events, there can only really be a single reason for an event occuring.

### DIFF
--- a/docs/http_api.rst
+++ b/docs/http_api.rst
@@ -291,10 +291,6 @@ Status events ``POST``\ed to the ``status_url`` specified in :http:post:`/channe
        The UUID of the channel the status event occurred for.
    :param str status:
        The :ref:`status level <status-levels>` this event was categorised under.
-   :param list reasons:
-       An array of objects describing the reasons for the status event, where
-       each object contains a ``message`` field with a human-readable string
-       value and a ``type`` field with a programmatically usable string value.
    :param str type:
        A programmatically usable string value describing the reason for the
        status event.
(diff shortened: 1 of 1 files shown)


justinvdm:
Ready for review.

docs/http_api.rst
```diff
@@ -309,10 +315,8 @@ Status events ``POST``\ed to the ``status_url`` specified in :http:post:`/channe
       status: "major",
       component: "smpp",
       channel_id: "channel-uuid-5678",
-      reasons: [{
-        type: "connection_lost",
-        message: "Connection lost"
-      }],
+      type: "connection_lost",
+      message: "Connection lost",
```
justinvdm: This change does mean that `type` and `message` are now mandatory fields for things publishing status events, but I think that is maybe a reasonable thing to enforce.

docs/http_api.rst
```diff
@@ -295,6 +295,12 @@ Status events ``POST``\ed to the ``status_url`` specified in :http:post:`/channe
        An array of objects describing the reasons for the status event, where
        each object contains a ``message`` field with a human-readable string
        value and a ``type`` field with a programmatically usable string value.
+   :param str type:
+       A programmatically usable string value describing the reason for the
+       status event.
+   :param str message:
+       A human-readable string value describing the reason for the status
+       event.
```
rudigiesler: We probably also want to remove `reasons` from this specification

justinvdm: Woops, forgot to remove it, thanks.

rudigiesler:
:+1:

PR #43, opened 2015-10-16
Add link to Rudi's video.

hodgestar:
We should add a link to the video to the documentation.


hodgestar:
Ready for review.

Is this in the correct place in the docs?

rudigiesler:
I think this is a good place for now. We must just remember to update it when we create new videos for new versions.

rudigiesler:
:+1:

PR #48, opened 2015-10-21
Support GET health events

rudigiesler:
A GET request to `/channels/<channel-id>` should have the latest health event from each component in it.

### DIFF
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -139,4 +139,7 @@ class StatusStore(BaseStore):
         dictionary'''
         key = self.get_key(channel_id)
         statuses = yield self.load_all(key)
-        returnValue({k: json.loads(v) for k, v in statuses.iteritems()})
+        res = {}
+        for k, v in statuses.iteritems():
+            res[k] = json.loads(v)
+        returnValue(res)
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -1,3 +1,4 @@
+import json
 from twisted.internet.defer import inlineCallbacks, returnValue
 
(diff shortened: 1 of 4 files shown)


rudigiesler:
Ready for review.

junebug/stores.py
```diff
@@ -131,3 +132,14 @@ def store_status(self, channel_id, status):
...
+
+    @inlineCallbacks
+    def get_statuses_dict(self, channel_id):
+        '''Returns the latest status message for each component in a
+        dictionary'''
+        key = self.get_key(channel_id)
+        statuses = yield self.load_all(key)
+        res = {}
+        for k, v in statuses.iteritems():
+            res[k] = json.loads(v)
```
justinvdm: #46 adds a [`api_from_status`](https://github.com/praekelt/junebug/blob/feature/issue-46-post-received-events-to-status-url-in-channel-status-worker/junebug/utils.py#L97-L105) utility. I think it would help for us to use that here for consistency and so that we aren't passing through undocumented message fields only relevant to vumi (like `helper_metadata`).

rudigiesler:
Just realised that this PR will also need an `api_from_status` function, so will wait for https://github.com/praekelt/junebug/pull/46/files#diff-d0301fcb8032d0b85f3f32d0bbccfd6cR97 to land before updating it to use the function.

justinvdm:
:+1: once we are using `api_from_status`.

rudigiesler:
So there was a slight architectural change that was made with how the status store returns items, as well as now using the `api_from_status` method. Ready for another review.

justinvdm:
:+1:

PR #44, opened 2015-10-19
Create status receive worker

rudigiesler:
Create a vumi worker that will receive status messages related to a transport.

PR #46, opened 2015-10-21
POST received events to status_url in channel status worker

justinvdm:
#44 adds a worker that consumes status events published by a transport. The plan is to change the worker to post received events to a `status_url` given to it via its config.

### DIFF
--- a/junebug/tests/helpers.py
+++ b/junebug/tests/helpers.py
@@ -203,9 +203,9 @@ class JunebugTestBase(TestCase):
             'vumi', queue)
 
     def assert_was_logged(self, msg):
-        return any(
+        return self.assertTrue(any(
             msg in log.getMessage()
-            for log in self.logging_handler.buffer)
+            for log in self.logging_handler.buffer))
 
     def assert_request(self, req, method=None, body=None, headers=None):
         if method is not None:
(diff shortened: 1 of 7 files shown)


justinvdm:
Ready for review.

justinvdm:
So, the `assert_was_logged` appears to have been giving us false positives, fixing, not ready for review yet.

junebug/tests/helpers.py
```diff
@@ -161,3 +201,28 @@ def get_dispatched_messages(self, queue):
...
+
+        if body is not None:
+            self.assertEqual(json.loads(req['body']), body)
+
+    def assert_body_contains(self, req, **fields):
+        body = json.loads(req['body'])
+
+        self.assertEqual(
+            dict((k, v) for k, v in body.iteritems()),
+            body)
```
rudigiesler: I'm a little confused as to what this is actually doing. Is it not just comparing `body` to itself? Should we be comparing it to `fields`?

justinvdm:
Ok, ready for review again.

junebug/tests/helpers.py
justinvdm: Ya, this looks wrong, should be comparing to fields, thanks, fixing.

justinvdm: Done.

rudigiesler:
:+1:

PR #19, opened 2015-10-22
Persistent state

rudigiesler:
We do store a list of all the channels in redis, but when junebug starts we don't start any of them. We should start all of the channels stored in redis when junebug starts up, so that they don't need to all be recreated every time the process restarts.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -1,3 +1,4 @@
+import json
 from klein import Klein
 import logging
 from werkzeug.exceptions import HTTPException
@@ -50,6 +51,17 @@ class JunebugApi(object):
 
         self.outbounds = OutboundMessageStore(
             self.redis, self.config.outbound_message_ttl)
+
+        yield self._start_stored_channels()
+
+    @inlineCallbacks
+    def _start_stored_channels(self):
+        ids = yield Channel.get_all(self.redis)
+        for id in ids:
+            properties = yield self.redis.get('%s:properties' % id)
+            properties = json.loads(properties)
+            channel = Channel(self.redis, self.config, properties, id)
+            yield channel.start(self.service)
 
     @inlineCallbacks
     def teardown(self):
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -1,4 +1,3 @@
-import json
 from klein import Klein
 import logging
 from werkzeug.exceptions import HTTPException
@@ -52,16 +51,7 @@ class JunebugApi(object):
         self.outbounds = OutboundMessageStore(
             self.redis, self.config.outbound_message_ttl)
 
-        yield self._start_stored_channels()
-
-    @inlineCallbacks
-    def _start_stored_channels(self):
-        ids = yield Channel.get_all(self.redis)
-        for id in ids:
-            properties = yield self.redis.get('%s:properties' % id)
-            properties = json.loads(properties)
-            channel = Channel(self.redis, self.config, properties, id)
-            yield channel.start(self.service)
+        yield Channel.start_all_channels(self.redis, self.config, self.service)
 
     @inlineCallbacks
     def teardown(self):
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -206,9 +206,7 @@ class Channel(object):
     def start_all_channels(cls, redis, config, parent):
         '''Ensures that all of the stored channels are running'''
         for id in (yield cls.get_all(redis)):
-            try:
-                yield cls.from_id(redis, config, id, parent)
-            except KeyError:
+            if id not in parent.namedServices:
                 properties = json.loads((
                     yield redis.get('%s:properties' % id)))
                 channel = cls(redis, config, properties, id)
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -201,6 +201,19 @@ class Channel(object):
         channels = yield redis.smembers('channels')
         returnValue(channels)
 
+    @classmethod
+    @inlineCallbacks
+    def start_all_channels(cls, redis, config, parent):
+        '''Ensures that all of the stored channels are running'''
+        for id in (yield cls.get_all(redis)):
+            try:
+                yield cls.from_id(redis, config, id, parent)
+            except KeyError:
+                properties = json.loads((
+                    yield redis.get('%s:properties' % id)))
+                channel = cls(redis, config, properties, id)
+                yield channel.start(parent)
+
     @inlineCallbacks
     def send_message(self, sender, outbounds, msg):
         '''Sends a message.'''
--- a/junebug/tests/helpers.py
+++ b/junebug/tests/helpers.py
@@ -164,6 +164,9 @@ class JunebugTestBase(TestCase):
     def start_server(self):
         '''Starts a junebug server. Stores the service to "self.service", and
         the url at "self.url"'''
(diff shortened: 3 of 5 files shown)


rudigiesler:
Ready for review, although please be critical. There are some things that I'm sure there are better ways of doing, but I couldn't come up with it.

junebug/api.py
```diff
@@ -51,6 +52,17 @@ def setup(self, redis=None, message_sender=None):
...
+        yield self._start_stored_channels()
+
+    @inlineCallbacks
+    def _start_stored_channels(self):
+        ids = yield Channel.get_all(self.redis)
+        for id in ids:
+            properties = yield self.redis.get('%s:properties' % id)
+            properties = json.loads(properties)
+            channel = Channel(self.redis, self.config, properties, id)
+            yield channel.start(self.service)
```
justinvdm: Would it maybe be worth encapsulating some of this in `Channel`? Maybe a classmethod that does something like `Channel.get_all_channels(self.redis, self.config)` that returns a list of `Channel` instances?

junebug/tests/helpers.py
```diff
@@ -182,10 +182,21 @@ def start_server(self):
...
 
+    @inlineCallbacks
+    def stop_server(self):
+        yield self.service.stopService()
+        for service in self.service.services:
+            yield service.stopService()
+        self.service.services = []
+        for service in self.service.namedServices.values():
+            yield service.stopService()
+        self.service.namedServices = {}
```
justinvdm: Would using `self.service.removeService()` on each child service maybe simplify this? It seems to handle stopping the service and removing it from `service.services` and `service.namedServices`.

justinvdm:
Left two minor comments, otherwise looks good.

junebug/tests/helpers.py
rudigiesler: Yes, this is what I was looking for, thank you.

rudigiesler:
Okay I've done some moving around of things, trying to use existing methods as much as possible. Ready for another review.

junebug/channel.py
```diff
@@ -201,6 +201,19 @@ def get_all(cls, redis):
         channels = yield redis.smembers('channels')
         returnValue(channels)
 
+    @classmethod
+    @inlineCallbacks
+    def start_all_channels(cls, redis, config, parent):
+        '''Ensures that all of the stored channels are running'''
+        for id in (yield cls.get_all(redis)):
+            try:
+                yield cls.from_id(redis, config, id, parent)
```
justinvdm: Would there possibly be cases where we have a channel in-memory but it is not started?

justinvdm: I think `channel.start` is a no-op for started channels (or should be), so maybe we should just have the `channel.start` apply to both cases (both in memory or created from redis).

junebug/tests/helpers.py
```diff
@@ -182,10 +182,19 @@ def start_server(self):
...
         addr = port.getHost()
         self.url = "http://%s:%s" % (addr.host, addr.port)
 
+    @inlineCallbacks
+    def stop_server(self):
+        yield self.service.stopService()
+        for service in self.service:
+            yield self.service.removeService(service)
+        for service in self.service.namedServices.values():
+            yield self.service.removeService(service)
```
justinvdm: I think the first loop should cover all services, or is it not so?

junebug/channel.py
rudigiesler: Having a look, currently `channel.start` is not a no-op for started channels, it will fail when we try to `setServiceParent` with `cannot have two services with the same name`. 

But I don't think there is a possible case when we have it in memory but not started. We use `disownServiceParent` when stopping, so even if it's in memory, it won't be able to be found.

junebug/tests/helpers.py
rudigiesler: I though so to, but the test fails without it, so the twisted docs are wrong

junebug/channel.py
justinvdm: I see. Cool, lets leave it as is. Not too important, but maybe as a future thing, we should consider making `channel.start` a no-op for started channels.

junebug/tests/helpers.py
jerith: From http://twistedmatrix.com/documents/current/api/twisted.application.service.IServiceCollection.html#removeService
> def removeService(service):
>   Remove a child service.
>   Only implementations of IService.disownServiceParent should use this method.

rudigiesler: @jerith Thanks. I'll change to use `disownServiceParent` instead.

justinvdm: Ah, then my suggestion to use `removeService` was probbaly a bad one. @jerith Suggestions on a different way to remove the service?

justinvdm: Ah, forgot about `disownServiceParent`, +1.

junebug/tests/helpers.py
```diff
@@ -182,10 +182,19 @@ def start_server(self):
...
         addr = port.getHost()
         self.url = "http://%s:%s" % (addr.host, addr.port)
 
+    @inlineCallbacks
+    def stop_server(self):
+        yield self.service.stopService()
+        for service in self.service:
+            service.disownServiceParent()
+        for service in self.service.namedServices.values():
+            service.disownServiceParent()
```
justinvdm: Do we maybe now not need the second loop?

justinvdm:
Left one tiny comment, otherwise :+1: from me.

junebug/tests/helpers.py
rudigiesler: No, it seems like we still need it. I think that the first one just goes through unnamed services. You need to get the named services separately. I'm just not sure on the best way to get the named services.

justinvdm: Ah ok cool, thanks, happy leaving it as is.

justinvdm:
:+1: from me

rudigiesler:
@justinvdm Thanks. I'll wait for the review from @jerith tomorrow before landing.

jerith:
Why does `JunebugTestBase.start_server()` not actually start the `JunebugService` object it creates?

By skipping that, you're not setting up the underlying service machinery properly. This is likely the cause of child services not starting and stopping properly. (It's also probably going to cause problems in the future when the test setup and real code diverge)

jerith:
Related to the above, the tests also don't seem to ever check if the thing they're supposed to have started is actually running.

rudigiesler:
@jerith The reason for this is because we don't actually want to start the service, because that would involve connecting to real AMQP and real Redis. So we 'start' it manually, and substitute our fakeAMQP and fakeRedis

jerith:
The consequence of the current mechanism is that your services behave differently in tests and production, which significantly reduces the value of the tests.

Instead, you need a way to inject the fake AMQP and Redis into the normal setup process and then call that.

rudigiesler:
@jerith I've created an issue https://github.com/praekelt/junebug/issues/50 to address your concerns, and added a comment to the messy teardown code to explain things. Care to give it another look?

jerith:
@rudigiesler: Perhaps add a similar comment to the setup side of the test machinery as well?

:+1: either way if you've done some manual testing with the non-test startup code and are happy that it works.

PR #51, opened 2015-10-29
Create base class for junebug plugins

rudigiesler:


junebug/plugin.py
```diff
@@ -0,0 +1,23 @@
+class JunebugPlugin(object):
+    '''Base class for all Junebug plugins'''
+
+    def start_plugin(self, config):
+        '''Can be overridden with any required startup code for the plugin.
+        Can return a deferred.
+
+        config - Config that Junebug was started with.'''
```
justinvdm: Should we maybe do this in the sphinx `:param:` format here so we can autogen? Not too important though.

rudigiesler: Yes we probably should. I'll change that

rudigiesler:
I'm not sure of how I should limit the scope of this PR? Should we end here, or should we go on an start adding the places in the API where the plugins should be instantiated and called?

justinvdm:
Minor comment, otherwise looks great.

justinvdm:
> I'm not sure of how I should limit the scope of this PR? Should we end here, or should we go on an start adding the places in the API where the plugins should be instantiated and called?

I vote we end here and add the places in the api where the plugins should be instantiated and called in a separate PR, once this PR lands we can probably work on things in parallel.

justinvdm:
:+1:

PR #53, opened 2015-10-30
Add infrastructure to use plugins

rudigiesler:
In https://github.com/praekelt/junebug/pull/51 we added the interface for Junebug plugins. We now need to add infrastructure to configure which plugins get started when Junebug starts, and to have all those plugins be called in the correct places.


justinvdm:
:+1:

PR #52, opened 2015-11-02
Add an nginx junebug plugin

justinvdm:
For smaller projects using junebug, it would be helpful to have a plugin that maintains a set of nginx config files, where each file corresponds to an http-based channel and contains a virtual server configuration for that channel that routes requests to the relevant port that the channel is listening on.

### DIFF
--- a/junebug/plugins/nginx/location.template
+++ b/junebug/plugins/nginx/location.template
@@ -1,3 +1,3 @@
 location %(external_path)s {
-  proxy_pass %(internal_url)s
+  proxy_pass %(internal_url)s;
 }
--- a/junebug/plugins/nginx/location.template
+++ b/junebug/plugins/nginx/location.template
@@ -1,0 +1,3 @@
+location %(external_path)s {
+  proxy_pass %(internal_url)s
+}
--- a/junebug/plugins/nginx/plugin.py
+++ b/junebug/plugins/nginx/plugin.py
@@ -1,9 +1,15 @@
 from os import path
+from distutils.dir_util import mkpath
+from urlparse import urljoin
 
 from confmodel import Config
 from confmodel.fields import ConfigText
 
 from junebug.plugin import JunebugPlugin
+from junebug.utils import channel_public_http_properties
+
+
+DIRNAME = path.dirname(__file__,)
 
 
 class NginxPluginConfig(Config):
@@ -22,8 +28,12 @@ class NginxPluginConfig(Config):
         required=True, static=True)
 
     vhost_template = ConfigText(
-        "Path to the template file to use for the vhost file",
-        default='%s/vhost.template' % (path.dirname(__file__,)), static=True)
+        "Path to the template file to use for the vhost config",
+        default='%s/vhost.template' % (DIRNAME,), static=True)
+
+    location_template = ConfigText(
+        "Path to the template file to use for each channel's location config",
+        default='%s/location.template' % (DIRNAME,), static=True)
 
 
 class NginxPlugin(JunebugPlugin):
@@ -35,7 +45,18 @@ class NginxPlugin(JunebugPlugin):
     def start_plugin(self, config, junebug_config):
         self.config = NginxPluginConfig(config)
         self.vhost_template = read(self.config.vhost_template)
+        self.location_template = read(self.config.location_template)
         write(self.config.vhost_file, self.get_vhost_config())
+
+    def channel_started(self, channel):
+        properties = channel_public_http_properties(channel._properties)
+
+        if properties is not None and properties['enabled']:
+            mkpath(self.config.locations_dir)
+
+            write(
+                self.get_location_path(channel.id),
+                self.get_location_config(properties))
 
     def get_vhost_config(self):
         return self.vhost_template % self.get_vhost_context()
@@ -43,8 +64,23 @@ class NginxPlugin(JunebugPlugin):
     def get_vhost_context(self):
         return {
             'server_name': self.config.server_name,
-            'includes': path.join(self.config.locations_dir, '*.conf'),
+            'includes': path.join(self.config.locations_dir, '*.conf')
         }
+
+    def get_location_config(self, properties):
+        return self.location_template % self.get_location_context(properties)
+
+    def get_location_context(self, properties):
+        web_path = properties['web_path']
+        base_url = 'http://localhost:%s' % (properties['web_port'],)
+
+        return {
+            'external_path': web_path,
+            'internal_url': urljoin(base_url, web_path)
+        }
+
+    def get_location_path(self, id):
+        return path.join(self.config.locations_dir, "%s.conf" % (id,))
 
 
 def read(filename):
--- a/junebug/plugins/nginx/plugin.py
+++ b/junebug/plugins/nginx/plugin.py
@@ -106,7 +106,7 @@ class NginxPlugin(JunebugPlugin):
 
 def reload_nginx():
     if in_path('nginx'):
-        subprocess.call(['nginx', '-s', 'reload'])
+        subprocess.check_call(['nginx', '-s', 'reload'])
     else:
         log.error('Cannot reload nginx, nginx not found in path')
 
--- a/junebug/plugins/nginx/plugin.py
+++ b/junebug/plugins/nginx/plugin.py
@@ -3,6 +3,7 @@ import subprocess
 from os import path, remove
 from distutils.dir_util import mkpath
 from urlparse import urljoin
+from pkg_resources import resource_filename
 
 from confmodel import Config
 from confmodel.fields import ConfigText
@@ -12,7 +13,10 @@ from junebug.utils import channel_public_http_properties
 
 
 log = logging.getLogger(__name__)
-DIRNAME = path.dirname(__file__)
+
+
+def resource_path(filename):
+    return resource_filename('junebug.plugins.nginx', filename)
 
 
 class NginxPluginConfig(Config):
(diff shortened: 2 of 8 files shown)


junebug/plugins/nginx/vhost.template
```diff
@@ -0,0 +1,4 @@
+server {
+  server_name %(server_name)s;
+  include %(includes)s;
+}
```
justinvdm: Any other settings that should be in the template?

junebug/plugins/nginx/location.template
```diff
@@ -0,0 +1,3 @@
+location %(external_path)s {
+  proxy_pass %(internal_url)s;
+}
```
justinvdm: Any other settings that should be in the location config template?

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+class NginxPluginConfig(Config):
+    '''Config for :class:`NginxJunebugPlugin`'''
+
+    vhost_file = ConfigText(
+        "The file to write the junebug nginx vhost file to",
+        default='/etc/nginx/sites-enabled/junebug.conf', static=True)
+
+    locations_dir = ConfigText(
+        "The directory to write location block config files to",
+        default='/etc/nginx/includes/junebug/', static=True)
```
justinvdm: We currently recursively create dirs that don't exist here. The docs suggest the created dir's permissions are `0777`, but it seems to create the dirs with `0755`. Maybe that is an okay permission level for now, and we should think about a configurable level as a future thing?

junebug/tests/test_utils.py
```diff
@@ -298,3 +299,99 @@ def test_api_from_status(self):
...
         })
+
+    def test_public_http_properties_explicit(self):
+        result = channel_public_http_properties({
+            'web_path': '/baz/quux',
+            'web_port': 2121,
+            'public_http': {
+                'web_path': '/foo/bar',
+                'web_port': 2323,
+            }
```
justinvdm: Some changes I made that is different to what we last discussed:
  - Using `public_http` instead of `nginx`, seems more descriptive of the purpose of this configuration, but happy to use a different name
  - The public http properties are now a sub-property of the channel config given to junebug via a `POST /channels/`, seems better placed there than outside of the config, but happy to move it outside of the config if it would make more sense.
   - Not sure if we discussed this, but we allow any of `web_path` and `web_port` to be included in the channel config, and the same field names in `public_http` just override those if present.

justinvdm:
Ready for review.

junebug/plugins/nginx/vhost.template
rudigiesler: There's usually a `listen 80;` here too, but I think if you leave it out it's the implicit default.

junebug/plugins/nginx/location.template
rudigiesler: Sometimes you set the headers `X-Real-IP`, `Host`, etc. But I don't think any of our transports care about those things; I think that this is okay.

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+
+    server_name = ConfigText(
+        "Server name to use for nginx vhost",
+        required=True, static=True)
+
+    vhost_template = ConfigText(
+        "Path to the template file to use for the vhost config",
+        default='%s/vhost.template' % (DIRNAME,), static=True)
```
rudigiesler: Is this so that people can set up their own auth certificates, passthrough to the Junebug API, etc? I'm wondering if it would be easier/better to allow a config string where you can put a text blob that would be included in the template, where you could place that sort of extra config?

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def start_plugin(self, config, junebug_config):
+        self.configured_channels = set()
+        self.config = NginxPluginConfig(config)
+        self.vhost_template = read(self.config.vhost_template)
+        self.location_template = read(self.config.location_template)
+        write(self.config.vhost_file, self.get_vhost_config())
+
+    def stop_plugin(self):
+        ensure_removed(self.config.vhost_file)
```
rudigiesler: Should we not also remove all of the configured channels? If not all of the same channels get started up when the plugin gets started again, then we will have some locations leading to nowhere.

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+            'server_name': self.config.server_name,
+            'includes': path.join(self.config.locations_dir, '*.conf')
+        }
+
+    def get_location_config(self, properties):
+        return self.location_template % self.get_location_context(properties)
+
+    def get_location_context(self, properties):
+        web_path = properties['web_path']
+        base_url = 'http://localhost:%s' % (properties['web_port'],)
```
rudigiesler: This might be rabbitholing and maybe we can leave this for a future thing, but what if I don't know what ports are open on the server, and I don't want to have to check, so I just want to be able to set `web_port` to `0`, and have an available port chosen for me (this could be useful if a web UI is creating the channel, it would be strange to ask the user for an open port on your server). We would then probably need to handle this somehow. Probably not for this PR, but just a thought for the future.

junebug/plugins/nginx/plugin.py
hodgestar: I wonder what happens if Junebug doesn't have permission to write to `/etc/nginx`?

rudigiesler: I think that's okay for now.

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+        "Server name to use for nginx vhost",
+        required=True, static=True)
+
+    vhost_template = ConfigText(
+        "Path to the template file to use for the vhost config",
+        default='%s/vhost.template' % (DIRNAME,), static=True)
+
+    location_template = ConfigText(
+        "Path to the template file to use for each channel's location config",
+        default='%s/location.template' % (DIRNAME,), static=True)
```
hodgestar: I wonder if we should be using `pkg_resources.resource_filename` here instead of `DIRNAME`? (there are situations in which a `__file__` may not point to a real file (e.g. it may be inside a zip import), but perhaps that isn't important here).

rudigiesler:
Just one main comment about cleaning up when shutting down the plugin. The rest are mainly questions/answers.

junebug/plugins/nginx/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+            'internal_url': urljoin(base_url, web_path)
+        }
+
+    def get_location_path(self, id):
+        return path.join(self.config.locations_dir, "%s.conf" % (id,))
+
+
+def reload_nginx():
+    if in_path('nginx'):
+        subprocess.call(['nginx', '-s', 'reload'])
```
hodgestar: Do we maybe want to use `check_call` or `check_output` instead of `.call`?

junebug/plugins/nginx/vhost.template
hodgestar: Explicit better than implicit?

junebug/tests/test_utils.py
hodgestar: Would be nice to be able to automatically assign ports or to use named Unix sockets instead of TCP? Maybe something for a different PR.

junebug/plugins/nginx/plugin.py
justinvdm: @hodgestar I think we'll get a:
```
distutils.errors.DistutilsFileError: could not create 'foo/bar': Permission denied
```

Maybe that is okay? Not sure there's much that can be done about it.

junebug/plugins/nginx/plugin.py
justinvdm: Hmm, good point. Agreed that it isn't in this PR's scope, but we should discuss this.

junebug/plugins/nginx/plugin.py
justinvdm: We can, how are they better than `call` here?

junebug/plugins/nginx/plugin.py
justinvdm: Yep, this is for any vhost configuration people might want to do. I'm not sure if a text blob in a config field is a better way of approaching this though. A template of the file seems closer to the way this kind of thing is usually done with configuration management things like puppet. Any particular reason you think it would be easier?

junebug/plugins/nginx/plugin.py
justinvdm: Doh, ofc, thanks :)

junebug/plugins/nginx/plugin.py
hodgestar: They raise errors if a non-zero return code is returned?

justinvdm: Ah, +1, thanks.

junebug/tests/test_utils.py
justinvdm: Good point, also [raised by @rudigiesler](https://github.com/praekelt/junebug/pull/52#discussion_r43733596). Lets discuss it in slack?

justinvdm:
Comments either replied to or resolved, ready for another review I think.

junebug/plugins/nginx/plugin.py
hodgestar: I had a similar thought when I read through the document. Maybe the `web_path` can default to something based on the channel id and the `web_port` can eventually default to a Unix socket? Definitely for a later PR though.

junebug/plugins/nginx/plugin.py
rudigiesler: With the text blob, you don't need to know how the default config looks to make your own config. Also, you can't break things by forgetting to add the `include` for junebug things.

hodgestar: I'm for the template-style thing we have currently. If we cobble it
together from bits and pieces, there'll be sections the person using
Junebug can't control (e.g. the include might need to be inside some other
block for silly reasons).

junebug/plugins/nginx/plugin.py
justinvdm: #55 created.

junebug/tests/test_utils.py
justinvdm: #55 created.

rudigiesler:
:+1:

PR #57, opened 2015-11-04
Document nginx plugin

justinvdm:


junebug/plugins/nginx/plugin.py
```diff
@@ -16,7 +16,7 @@
 
 
 def resource_path(filename):
-    return resource_filename('junebug.plugins.nginx', filename)
+    return path.join(path.dirname(__file__), filename)
```
justinvdm: Building the docs wasn't working with `pkg_resources.resource_filename`, was getting:

```
  File "/.../junebug/ve/lib/python2.6/site-packages/pkg_resources.py", line 1547, in _extract_resource
    timestamp, size = self._get_date_and_size(self.zipinfo[zip_path])
KeyError: 'junebug/plugins/nginx/vhost.template'
```

Guessing `resource_filename` wasn't happy with junebug being installed as an egg, but really not too sure.

justinvdm:
Ready for review.

junebug/plugins/nginx/plugin.py
rudigiesler: We probably also want to remove the import for `resource_filename` then.

hodgestar: We could also set `zip_safe=False` in `setup.py`, I think?

justinvdm: Done. I thought the reason we used `pkg_resources.resource_filename` was because the `__file__` approach wouldn't work in a zip, though.

justinvdm:
Ready for re-review.

rudigiesler:
:+1: from me

hodgestar:
:+1:

PR #60, opened 2015-11-09
Create message rate store

rudigiesler:
Create a store that can get called whenever the counter should be incremented, and be able to get the current msgs/s, with configurable bucket sizes.

### DIFF
--- a/junebug/stores.py
+++ b/junebug/stores.py
@@ -14,6 +14,8 @@ class BaseStore(object):
     :type ttl: integer
     '''
 
+    USE_DEFAULT_TTL = object()
+
     def __init__(self, redis, ttl=None):
         self.redis = redis
         self.ttl = ttl
@@ -21,7 +23,7 @@ class BaseStore(object):
     @inlineCallbacks
     def _redis_op(self, func, id, *args, **kwargs):
         ttl = kwargs.pop('ttl')
-        if ttl is None:
+        if ttl is self.USE_DEFAULT_TTL:
             ttl = self.ttl
         val = yield func(id, *args, **kwargs)
         if ttl is not None:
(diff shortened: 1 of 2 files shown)


rudigiesler:
@hodgestar Ready for review.

junebug/stores.py
```diff
@@ -142,3 +151,46 @@ def get_statuses(self, channel_id):
...
+        return self.get_key(channel_id, label, bucket)
+
+    def increment(self, channel_id, label, bucket_size):
+        '''Increments the correct counter. Should be called whenever a message
+        that should be counted is received.
+
+        Note: bucket_size should be kept constant for each channel_id and label
+        combination. Changing bucket sizes results in undefined behaviour.'''
+        key = self._get_current_key(channel_id, label, bucket_size)
+        self.ttl = bucket_size * 2
```
hodgestar: Should `bucket_size` not be passed to the `MessageRateStore` constructor? That would avoid having to pass it in to the individual methods and prevent people from passing in different values?

junebug/stores.py
```diff
@@ -142,3 +151,46 @@ def get_statuses(self, channel_id):
...
+        return self.increment_id(key)
+
+    @inlineCallbacks
+    def get_messages_per_second(self, channel_id, label, bucket_size):
+        '''Gets the current message rate in messages per second.
+
+        Note: bucket_size should be kept constant for each channel_id and label
+        combination. Changing bucket sizes results in undefined behaviour.'''
+        key = self._get_last_key(channel_id, label, bucket_size)
+        self.ttl = None
```
hodgestar: I feel that juggling `self.ttl` here is not really worth it. If a few extra redis metric keys live for a bit longer, it's not a much of an issue, but if all metric keys don't have expirey set because we forget to set the TTL back to a value, that's a problem?

junebug/stores.py
hodgestar: If we pass the bucket size into the constructor, we can also set the ttl there.

hodgestar:
Left some questions about bucket size and TTL setting, but otherwise looks good.

junebug/stores.py
rudigiesler: I originally wanted to inherit as much as possible from the `BaseStore`, but I see now that it is hurting us more than helping us.

The reason I wanted `bucket_size` to be passed per function is that in the API when we're fetching the msgs/s, `inbound` and `outbound` could be configured to have different bucket sizes. This reason is why we had `channel_id` being passed to every function for the other stores, instead of setting the channel id in the constructor.

The reason that `ttl` is being set here, is because we want to set the ttl when we increment, but we don't want to set the ttl again when we fetch the value (there we set `ttl` to `None`, so that we don't call `EXPIRE`.

hodgestar: Maybe we need a `ttl` flag that can override `self.ttl`? Or a `reset_ttl=False` flag?

rudigiesler:
@hodgestar I've added a `ttl` keyword argument that can be used to override the TTL value. Thoughts?

junebug/stores.py
```diff
@@ -19,32 +20,44 @@ def __init__(self, redis, ttl=None):
 
     @inlineCallbacks
     def _redis_op(self, func, id, *args, **kwargs):
+        ttl = kwargs.pop('ttl')
+        if ttl is None:
+            ttl = self.ttl
```
hodgestar: If `self.ttl = 5`, setting `ttl = None` in the kwargs won't override the `5`? Just checking whether that is intentional?

rudigiesler: Yes this is intentional. `None` means to use the default (and is the default for all of the methods) that is set when you create the class. I'm not sure if it's possible to make this clearer, maybe if the argument doesn't exist, it should use the default, and if it does exist, it uses that value? It would make things a bit more complicated though. Thoughts?

hodgestar: I think to make this work we'd need to use an explicit sentinel value, e.g.

```
USE_DEFAULT_TTL = object()

def _redis_op(...):
    ttl = kwargs.pop('ttl')
    if ttl is self.USE_DEFAULT_TTL:
        ttl = self.ttl

def get_id(..., ttl=USE_DEFAULT_TTL):
    ...
```

Too complicated?

rudigiesler: I think we can do that, yes. Looks good.

rudigiesler:
@hodgestar Changes made, ready for another review.

hodgestar:
:+1:

PR #62, opened 2015-11-09
Store message rates for MO messages

rudigiesler:
We need to track the message rates for MO messages. This will be done in the message forwarding worker.

### DIFF
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -259,6 +259,7 @@ class Channel(object):
             'redis_manager': self.config.redis,
             'inbound_ttl': self.config.inbound_message_ttl,
             'outbound_ttl': self.config.outbound_message_ttl,
+            'message_rate_bucket': self.config.inbound_message_rate_bucket,
         }
 
     @property
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -259,7 +259,7 @@ class Channel(object):
             'redis_manager': self.config.redis,
             'inbound_ttl': self.config.inbound_message_ttl,
             'outbound_ttl': self.config.outbound_message_ttl,
-            'message_rate_bucket': self.config.inbound_message_rate_bucket,
+            'metric_window': self.config.metric_window,
         }
 
     @property
--- a/junebug/command_line.py
+++ b/junebug/command_line.py
@@ -83,9 +83,9 @@ def create_parser():
         'plugin config. Must contain a `type` key, with the full python class '
         'path of the plugin')
     parser.add_argument(
-        '--inbound-message-rate-bucket', '-ibucket', type=float,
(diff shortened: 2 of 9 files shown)


rudigiesler:
Relies on #60

junebug/channel.py
```diff
@@ -259,6 +259,7 @@ def _application_config(self):
             'redis_manager': self.config.redis,
             'inbound_ttl': self.config.inbound_message_ttl,
             'outbound_ttl': self.config.outbound_message_ttl,
+            'message_rate_bucket': self.config.inbound_message_rate_bucket,
```
hodgestar: Should the config variable really be `inbound_message_rate_bucket`? I'm a bit confuse by the `inbound_` prefix on the one name and not on the other. Also wondering about the name of this option -- `message_rate_bucket` seems to imply the name of a bucket, rather than the size of a bucket.

Wondering what a better name is. Maybe `message_rate_window`? Or `metric_window`?

rudigiesler: `window` is great, that's the word I was looking for. The reason the `inbound` is missing here is because it's for the message forwarding worker, which only deals with inbound messages, happy to change it though.

hodgestar: I would vote for just `message_rate_window` or `metric_window` everywhere, unless there is a reason to have separate window sizes?

rudigiesler: I can't think of any reason to have it separate, so I think I'll change it to `metric_window`

rudigiesler:
@hodgestar Ready for review.

junebug/command_line.py
```diff
@@ -82,6 +82,10 @@ def create_parser():
         help='Add a plugins to the list of plugins, as a json blob of the '
         'plugin config. Must contain a `type` key, with the full python class '
         'path of the plugin')
+    parser.add_argument(
+        '--metric-window', '-mw', type=float,
+        dest='metric_window', help='The size of each bucket '
+        '(in seconds) to use for metrics. Defaults to 1 second.')
```
hodgestar: Is 1 second a good default?

hodgestar:
Left one question, but otherwise looks good.

junebug/command_line.py
rudigiesler: I'm not sure, what do we use in other places? The window really depends on how fast you're sending messages, the faster the shorter you want the window to be.

hodgestar: Maybe 10s? I guess it's a balance between responsiveness and showing either
0 or 1 all the time.

rudigiesler:
@hodgestar Default metric window changed to 10 seconds. Ready for another review.

hodgestar:
:+1:

PR #64, opened 2015-11-11
Add message rates to channel status

rudigiesler:
When a GET request is sent to `/channels/<channel-id>`, the inbound and outbound message rates should be returned.

### DIFF
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -8,7 +8,7 @@ from vumi.message import TransportUserMessage
 from vumi.service import WorkerCreator
 from vumi.servicemaker import VumiOptions
 
-from junebug.stores import StatusStore
+from junebug.stores import StatusStore, MessageRateStore
 from junebug.utils import api_from_message, message_from_api, api_from_status
 from junebug.error import JunebugError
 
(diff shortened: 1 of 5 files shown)


junebug/tests/helpers.py
```diff
@@ -257,6 +257,22 @@ def assert_body_contains(self, req, **fields):
...
+            self, level=None, components={}, inbound_message_rate=0,
+            outbound_message_rate=0):
+        '''Generates a status that the http API would respond with, given the
+        same parameters'''
+        return {
+            'status': level,
+            'components': components,
+            'inbound_message_rate': inbound_message_rate,
+            'outbound_message_rate': outbound_message_rate,
+        }
```
rudigiesler: My thinking here is that when we add more fields, or change how the fields are organised, we can just change this function (with relevant defaults), instead of changing all of the relevant tests.

rudigiesler:
Ready for review.

junebug/channel.py
```diff
@@ -178,11 +184,15 @@ def _get_status(self):
                 key=status_values.get)
         except ValueError:
             # No statuses
-            returnValue({})
+            status = None
```
hodgestar: I'm wondering whether `None` is the most sensible return value here, but I think it is. Just flagging this for thinking about.

junebug/tests/helpers.py
hodgestar: Woot.

hodgestar:
:+1:

Left one comment about whether `None` is the most sensible return value for the global status when there are no components, but I think it is reasonable.

PR #66, opened 2015-11-13
Add logging service

rudigiesler:
Create a logging service similar to `vumi.sentry.SentryLoggerService` that logs to a file instead of sentry, using `twisted.python.logfile` to assist with logfile rotation.

### DIFF
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -1,0 +1,127 @@
+import json
+import logging
+import time
+
+from twisted.python import log
+from twisted.python.log import ILogObserver
+from twisted.python.logfile import LogFile
+from twisted.application.service import Service
+
+from zope.interface import implements
+
+DEFAULT_LOG_CONTEXT_SENTINEL = "_JUNEBUG_CONTEXT_"
+
+
+class JunebugLogObserver(object):
+    """Twisted log observer that logs to a rotated log file."""
+    implements(ILogObserver)
+
+    DEFAULT_ERROR_LEVEL = logging.ERROR
+    DEFAULT_LOG_LEVEL = logging.INFO
+    LOG_LEVEL_THRESHOLD = logging.INFO
+    LOG_ENTRY = '%[(timestamp)s] '
+
+    def __init__(self, logfile, worker_id, log_context_sentinel=None):
+        '''
+        Create a new JunebugLogObserver.
+
+        :param logfile: File to write logs to.
+        :type logfile: :class:`twisted.python.logfile.LogFile`
+        :param str worker_id: ID of the worker that the log is for.
+        '''
+        if log_context_sentinel is None:
+            log_context_sentinel = DEFAULT_LOG_CONTEXT_SENTINEL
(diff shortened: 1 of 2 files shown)


rudigiesler:
Ready for review.

junebug/logging_service.py
```diff
@@ -0,0 +1,127 @@
...
+    def stopService(self):
+        if self.running:
+            self.logger.removeObserver(self.log_observer)
+            self.logfile.close()
+        return super(JunebugLoggerService, self).stopService()
+
+    def registered(self):
+        return (
+            getattr(self, 'log_observer', None) and
+            self.log_observer in self.logger.observers)
```
hodgestar: A common trick is to set `log_observer = None` on the class itself, so that if it hasn't been defined on the instance it can still be accessed.

junebug/tests/test_logging_service.py
```diff
@@ -0,0 +1,197 @@
...
+        '''A failure should be logged with the correct format.'''
+        e = ValueError("foo error")
+        f = Failure(e)
+        self.obs({
+            'failure': f, 'system': 'foo', 'isError': 1,
+            'message': [e.message]})
+
+        [log] = self.logfile.logs
+        log = json.loads(log)
+        log.pop('timestamp')
```
hodgestar: I can be nice to assert that the timestamp at least has the expected format.

junebug/tests/test_logging_service.py
```diff
@@ -0,0 +1,197 @@
...
+        self.assertEqual(logfile.worker_id, 'worker-id')
+        self.assertEqual(logfile.path, '/testpath/')
+        self.assertEqual(logfile.rotateLength, 1000000)
+        self.assertEqual(logfile.maxRotatedFiles, 7)
+
+    @inlineCallbacks
+    def test_logging(self):
+        '''The logging service should write logs to the logfile when the
+        service is running.'''
+        yield self.service.startService()
```
hodgestar: Should we also try log and message before the service is started and check that that doesn't log a message?

junebug/tests/test_logging_service.py
```diff
@@ -0,0 +1,197 @@
...
+        self.logger.msg("Hello", logLevel=logging.WARN)
+        [log] = logfile.logs
+        log = json.loads(log)
+        log.pop('timestamp')
+
+        self.assertEqual(log, {
+            'level': logging.WARN,
+            'logger': 'worker-id',
+            'message': 'Hello',
+        })
```
hodgestar: Maybe we could factor these extract log, pop timestamp, assert lines out into a helper function?

hodgestar:
Left some minor comments.

How are be planning to hook this up to the transports?

rudigiesler:
All the comments should be addressed. Hooking up to the transports will happen here: https://github.com/praekelt/junebug/issues/67 . The idea is to add the logging service as a child service to the transport worker.

hodgestar:
:+1:

PR #67, opened 2015-11-17
Start logger service with transport

rudigiesler:
When the transport is started, we should add the logger service to it (created in https://github.com/praekelt/junebug/pull/66) so that it can capture the logs of the worker.

### DIFF
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -270,6 +270,7 @@ class Channel(object):
         config = self._properties['config']
         config = self._convert_unicode(config)
         config['transport_name'] = self.id
+        config['worker_name'] = self.id
         config['publish_status'] = True
         return config
 
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -8,6 +8,7 @@ from vumi.message import TransportUserMessage
 from vumi.service import WorkerCreator
 from vumi.servicemaker import VumiOptions
 
+from junebug.logging_service import JunebugLoggerService
 from junebug.stores import StatusStore, MessageRateStore
 from junebug.utils import api_from_message, message_from_api, api_from_status
 from junebug.error import JunebugError
@@ -62,6 +63,7 @@ class Channel(object):
     STATUS_APPLICATION_ID = 'status:%s'
     APPLICATION_CLS_NAME = 'junebug.workers.MessageForwardingWorker'
     STATUS_APPLICATION_CLS_NAME = 'junebug.workers.ChannelStatusWorker'
+    JUNEBUG_LOGGING_SERVICE_CLS = JunebugLoggerService
 
     def __init__(self, redis_manager, config, properties, plugins=[], id=None):
         '''Creates a new channel. ``redis_manager`` is the redis manager, from
@@ -319,12 +321,17 @@ class Channel(object):
             transport_worker = self._create_transport()
 
         transport_worker.setName(self.id)
+
+        logging_service = self._create_junebug_logger_service()
+        transport_worker.addService(logging_service)
+
         transport_worker.setServiceParent(service)
         self.transport_worker = transport_worker
 
     def _start_application(self, service):
         worker = self._create_application()
         worker.setName(self.application_id)
+
         worker.setServiceParent(service)
         self.application_worker = worker
 
@@ -348,6 +355,11 @@ class Channel(object):
         return self._create_worker(
             self.STATUS_APPLICATION_CLS_NAME,
             self._status_application_config)
+
+    def _create_junebug_logger_service(self):
+        return JunebugLoggerService(
+            self.id, self.config.logging_path, self.config.log_rotate_size,
+            self.config.max_log_files)
 
     def _create_worker(self, cls_name, config):
         creator = WorkerCreator(self.options)
--- a/junebug/channel.py
+++ b/junebug/channel.py
@@ -358,7 +358,7 @@ class Channel(object):
             self._status_application_config)
 
     def _create_junebug_logger_service(self):
-        return JunebugLoggerService(
+        return self.JUNEBUG_LOGGING_SERVICE_CLS(
             self.id, self.config.logging_path, self.config.log_rotate_size,
             self.config.max_log_files)
 
--- a/junebug/command_line.py
+++ b/junebug/command_line.py
@@ -91,13 +91,13 @@ def create_parser():
         dest='logging_path', help='The path to place log files for each '
         'channel. Defaults to `logs/`')
     parser.add_argument(
-        '--log-rotate-size', '-lrs', type=str,
+        '--log-rotate-size', '-lrs', type=int,
         dest='log_rotate_size', help='The maximum size (in bytes) for each '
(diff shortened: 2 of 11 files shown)


rudigiesler:
Ready for review.

junebug/command_line.py
```diff
@@ -86,6 +86,18 @@ def create_parser():
...
+        dest='logging_path', help='The path to place log files for each '
+        'channel. Defaults to `logs/`')
+    parser.add_argument(
+        '--log-rotate-size', '-lrs', type=str,
+        dest='log_rotate_size', help='The maximum size (in bytes) for each '
+        'log file before it gets rotated. Defaults to 1000000.')
+    parser.add_argument(
+        '--max-log-files', '-mlf', type=str,
+        dest='max_log_files', help='The maximum number of log files to '
+        'keep before deleting old files. Defaults to unlimited log files.')
```
hodgestar: Maybe unlimited is a bad default? Perhaps let's go with 5?

junebug/logging_service.py
```diff
@@ -50,7 +50,7 @@ def logger_for_event(self, event):
         system = event.get('system', '-')
         parts = [self.worker_id]
         if system != '-':
-            parts.extend(system.split(','))
+            parts = system.split(',')
```
hodgestar: Not understanding why we're dropping the `worker_id` from `parts` here?

rudigiesler: The `system` has to have the `worker-_id` in it for us to process the log, so if we don't drop it, we end up with `worker_id.something_with_worker_id_in_it`, most of the time it will be `worker_id.worker_id`, which seemed unnecessary.

junebug/command_line.py
rudigiesler: The reason the default is `unlimited` is because we can't set it to `None` using the command line. Maybe we should add logic so that `0` is unlimited? Thoughts?

hodgestar: I like the idea of making `0` mean unlimited.

junebug/logging_service.py
hodgestar: Do we still need the `if system != -'` check then? And in what situations can system be `-`? Will all channel workers log these messages where `system` is `-`?

rudigiesler: No I guess we don't need that anymore. When `worker_id` is not in `system`, then it isn't logged. So the messages when `system` is `-` will mean that none of the channel workers will log it.

rudigiesler:
Changes made. Ready for review again.

junebug/channel.py
```diff
@@ -349,6 +357,11 @@ def _create_status_application(self):
             self.STATUS_APPLICATION_CLS_NAME,
             self._status_application_config)
 
+    def _create_junebug_logger_service(self):
+        return JunebugLoggerService(
```
rudigiesler: Yes it was, I'll change it to use the class variable

junebug/logging_service.py
```diff
@@ -72,13 +70,13 @@ def _log_to_file(self, event):
...
 
-        data['message'] = log.textFromEventDict(event)
-
         self.logfile.write(json.dumps(data))
 
     def __call__(self, event):
         if self.log_context_sentinel in event:
             return
+        if self.worker_id not in event.get('system', ''):
+            return
```
rudigiesler: I'm not sure that we would always expect the `<worker_id>` to be first in the system. What if we decide to organise it as `transports,<worker_id>` in future, to group all the transports, if we wanted to also return say application logs as well as transport logs in the future, because then we would also want `applications,<worker_id>`. 

I'm not sure I understand about the `.` after the worker id? For our current case, the `system` would just be `<worker_id>`, no `.`.

hodgestar: Would `self.worker_id not in event.get('system', '').split(',')` be a
better check?

junebug/tests/helpers.py
```diff
@@ -143,15 +147,24 @@ def create_channel(
...
+                    'test_channel': transport_class
+                })
+
+        patched_type = properties['type']
+        properties['type'] = 'test_channel'
         channel = Channel(
             redis, config, properties, id=id, plugins=plugins)
-        properties['config']['transport_name'] = channel.id
         yield channel.start(self.service)
+        properties['type'] = patched_type
```
rudigiesler: We patch the `type` because we want the channel configuration (returned when `GET`ing the channel) to reflect a valid type, e.g. it should return `telnet` not `test_channel`.

junebug/command_line.py
```diff
@@ -140,7 +152,10 @@ def config_from_args(args):
     config['amqp'] = parse_amqp(config.get('amqp', {}), args)
     parse_channels(args)
     args['plugins'] = parse_plugins(config.get('plugins', []), args)
-    return JunebugConfig(conjoin(config, args))
+
+    combined = conjoin(config, args)
+    combined['max_log_files'] = combined.get('max_log_files') or None
```
rudigiesler: Done

junebug/tests/test_channel.py
```diff
@@ -55,12 +58,76 @@ def test_delete_channel(self):
...
+
+        logging_worker = worker.getServiceNamed('Junebug Worker Logger')
+        self.assertTrue(
+            isinstance(logging_worker, channel.JUNEBUG_LOGGING_SERVICE_CLS))
+
+    @inlineCallbacks
+    def test_start_channel_logging(self):
+        '''When the channel is started, the logging worker should be started
+        along with it.'''
+        self.patch(junebug.logging_service, 'LogFile', DummyLogFile)
```
rudigiesler: Done.

junebug/logging_service.py
rudigiesler: We could, I'm just not sure that I see a case where those two statements could be different, besides if there is a `,` in `worker_id`, which isn't possible.

hodgestar: The situation I'm worried about is if `worker_id` is `foo` and `system` is `foobar`. Then `worker_id` is in `system`, but not the way we want. It's very much a corner case. Sorry, should have made that clearer earlier.

rudigiesler: Never mind, I see the situation now. When we have a system `<worker_id>foo,bar`, then we probably don't want to log that.

I'll change it to that check.

rudigiesler:
I have responded and made changes to all of the comments. Ready for review again.

junebug/tests/helpers.py
hodgestar: What if we did:
```
if config is None:
    config = yield self.create_channel_config(channels={
        properties['type']: transport_class,
    })
```
Then the transport class would be used without having to mess around with the type in properties?

hodgestar:
Other than the config patching question, everything looks good.

junebug/tests/helpers.py
rudigiesler: Ah, that would work, and is much cleaner. Changing that now.

rudigiesler:
Config patching addressed. Ready for a (hopefully) final review.

hodgestar:
:+1: :maple_leaf:

PR #68, opened 2015-11-18
Create logging reading service

rudigiesler:
Create a service that will retrieve the last N logs, possibly across multiple files.

### DIFF
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -139,12 +139,10 @@ def reverse_read(filename, buf):
     '''
     with open(filename) as f:
         f.seek(0, os.SEEK_END)
-        total_size = remaining_size = f.tell()
-        offset = 0
+        remaining_size = f.tell()
         incomplete_line = None
         while remaining_size > 0:
-            offset = min(total_size, offset + buf)
-            f.seek(total_size - offset, os.SEEK_SET)
+            f.seek(max(0, remaining_size - buf), os.SEEK_SET)
             data = f.read(min(remaining_size, buf))
             # If we read no data, we should exit the loop
             if len(data) == 0:
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -143,13 +143,26 @@ def reverse_read(filename, buf):
         incomplete_line = None
         while remaining_size > 0:
             offset = min(total_size, offset + buf)
-            f.seek(-offset, os.SEEK_END)
+            f.seek(total_size - offset, os.SEEK_SET)
             data = f.read(min(remaining_size, buf))
-            remaining_size -= buf
+            # If we read no data, we should exit the loop
+            if len(data) == 0:
+                break
+            remaining_size -= len(data)
             lines = data.split('\n')
+            # If the last line of the file doesn't end in a new line, it is
+            # possible that the line hasn't been finished yet, so we should
+            # ignore that line.
+            if incomplete_line is None and data[-1] != '\n':
+                lines.pop(-1)
+            # If there is an incomplete line from the last read, we should
+            # append it to the end of the current read.
             if incomplete_line:
                 lines[-1] += incomplete_line
-            incomplete_line = lines.pop(0)
+            # If we still have any lines left, the first one could be
+            # incomplete, so store it for the next iteration.
+            if lines:
+                incomplete_line = lines.pop(0)
             for l in lines[::-1]:
                 if l is not '':
                     yield l
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -71,7 +71,7 @@ class JunebugLogObserver(object):
             data['instance'] = repr(failure.value)
             data['stack'] = failure.stack
 
-        self.logfile.write(json.dumps(data))
+        self.logfile.write(json.dumps(data) + '\n')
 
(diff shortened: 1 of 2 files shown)


rudigiesler:
Ready for review.

junebug/logging_service.py
```diff
@@ -122,3 +123,61 @@ def stopService(self):
...
+        string.
+    '''
+    with open(filename) as f:
+        f.seek(0, os.SEEK_END)
+        total_size = remaining_size = f.tell()
+        offset = 0
+        incomplete_line = None
+        while remaining_size > 0:
+            offset = min(total_size, offset + buf)
+            f.seek(-offset, os.SEEK_END)
```
hodgestar: I think this seek needs to use `os.SEEK_SET` because the file will be growing and thus the end of the file will be moving.

junebug/logging_service.py
```diff
@@ -122,3 +123,61 @@ def stopService(self):
...
+            data = f.read(min(remaining_size, buf))
+            remaining_size -= buf
```
hodgestar: This should be `-= len(data)` since the number of bytes requested might have been `remaining_size` and `.read()` may return fewer bytes than requested.

hodgestar: We might want to exit the loop if no data was read because that likely indicates that the file was truncated underneath us or that we have messed up our logic.

hodgestar: I wonder if we want to skip the last line of the file if it doesn't end in `\n` (i.e. is an incompletely written log line)?

junebug/logging_service.py
```diff
@@ -122,3 +123,61 @@ def stopService(self):
...
+    '''
+    logs = []
+
+    # Read the current log file
+    for line in reverse_read(logfile.path, buf):
+        logs.append(json.loads(line))
+        if len(logs) == lines:
+            return logs
+    # Read the rotated log files
+    for file_num in logfile.listLogs():
```
hodgestar: Maybe we can combine these two loops with `itertools.chain`? Although perhaps `logfile` should just have a message that lists all the log filenames in the right order?

hodgestar:
I seem to have left a lot of comments on `reverse_read`, but generally looking awesome.

junebug/logging_service.py
```diff
@@ -122,3 +123,61 @@ def stopService(self):
...
+        for line in reverse_read('%s.%d' % (logfile.path, file_num), buf):
+            logs.append(json.loads(line))
```
hodgestar: I wonder if we need to protect against non-JSON log lines?

junebug/logging_service.py
rudigiesler: twisted's `LogFile` does not have such a method, unfortunately, so we'll have to use `itertools.chain`

hodgestar: Or a small helper generator function that yields the filenames if that
seems clearer?

junebug/logging_service.py
rudigiesler: I thought about that too, but this is supposed to be used along with `JunebugLoggerService` (which is why it's in the same file), which only logs JSON log lines.

rudigiesler:
I've addressed the comments. Ready for another review.

junebug/logging_service.py
```diff
@@ -122,3 +124,72 @@ def stopService(self):
...
+        f.seek(0, os.SEEK_END)
+        total_size = remaining_size = f.tell()
+        offset = 0
+        incomplete_line = None
+        while remaining_size > 0:
+            offset = min(total_size, offset + buf)
+            f.seek(total_size - offset, os.SEEK_SET)
+            data = f.read(min(remaining_size, buf))
+            # If we read no data, we should exit the loop
+            if len(data) == 0:
```
hodgestar: So, a couple more thoughts around this:
* I think we should ditch the `total_size` -- we only ever care about `remaining_size` (i.e. where we last read from). Removing `total_size` should make the logic a bit clearer.
* There is a corner case we're not covering which is that if `f.read()` returns fewer bytes than requested, we won't have all the data up to where we last read from. I seem to remember that the blocking Python `.read` call makes multiple calls to the underlying C `fread` until the number of bytes requested have been returned or the file ends, but I don't have the Python source checked out right now. In short, I think we are fine as we are, but just wanted to mention the issue.

hodgestar:
Left one more comment but otherwise looks good.

junebug/logging_service.py
rudigiesler: We should be okay if `f.read()` returns fewer bytes than requested, since we're using `len(data)` for everything now.

I can see how we can simplify by removing `total_size` and `offset`, so I'll remove those.

rudigiesler: Actually, I see what you mean with the fewer bytes thing. Yes it could be a problem. We could maybe write our own read function that calls the `.read` function until we have the required amount of bytes. But if that's not a problem, then I think we can leave it for now.

rudigiesler:
I've made the changes in the last comment. Ready for review.

junebug/logging_service.py
hodgestar: Let's leave it for now.

hodgestar:
:+1:

PR #69, opened 2015-11-19
Create API endpoint for retrieving logs

rudigiesler:
Now that we store logs, and have created a reader to read the last `n` log entries, we should create an API endpoint that uses this to return the last `n` logs.


rudigiesler:
Ready for review.

docs/http_api.rst
```diff
@@ -165,6 +165,66 @@ Channels
...
 
+Logs
+^^^^
+
+.. http:get:: /channels/(channel_id:str)/logs
+
+   Get the most recent logs for a specific channel.
+
+   :query int n:
+       The amount of logs to fetch.
```
hodgestar: s/amount/number/ -- maybe a bikeshed? I think "amount" is for continuous quantities. We need a grammar expert. :)

docs/http_api.rst
```diff
@@ -165,6 +165,66 @@ Channels
...
+   :param str logger: The logger that created the log, usually the channel id.
+   :param int level:
+       The level of the logs. Corresponds to the levels found in the python
+       module :py:mod:`logging`.
+   :param float timestamp: Timestamp, in the format of seconds since the epoch.
+   :param str message: The message of the log.
+
+   In the case of an exception, the following fields may also be returned:
+
+   :param str class: The class of the exception.
```
hodgestar: I wonder if we should nest these inside an `exception` sub-dict? Just thinking that `class` and `instance` by themselves don't really indicate that they are related to an exception?

docs/http_api.rst
rudigiesler: Number sounds better :)

docs/http_api.rst
rudigiesler: Sounds good, I'll make the change.

docs/http_api.rst
```diff
@@ -165,6 +165,66 @@ Channels
...
+                level: 20,
+                timestamp: 987654321.0,
+                message: "Last log for the channel"
+            },
+            {
+                logger: "123-456-7a90",
+                level: 20,
+                timestamp: 987654320.0,
+                message: "Second last log for the channel"
+            }
```
hodgestar: Should we include an example of an exception log?

hodgestar: Should we include an example of an exception log?

rudigiesler: I added an example in https://github.com/praekelt/junebug/commit/d3cf551eae900f4c6036a50e5330a594d2617606

hodgestar:
I think we need `get_logs` and API tests for the `n=None` case?

hodgestar:
Other than the tests for `n=None`, looks good!

rudigiesler:
Changes made. Ready for another review.

hodgestar:
:+1:

PR #70, opened 2015-11-20
Bug fixes

rudigiesler:
Two bugs that were found in practice:

Redis ttl needs to be an int, so we need to round up and convert the float to an int.

If the system is `None`, we shouldn't have an exception, we should just not log anything.

### DIFF
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -79,7 +79,7 @@ class JunebugLogObserver(object):
     def __call__(self, event):
         if self.log_context_sentinel in event:
             return
-        if self.worker_id not in event.get('system', '').split(','):
+        if self.worker_id not in (event.get('system', '') or '').split(','):
             return
         log.callWithContext(self.log_context, self._log_to_file, event)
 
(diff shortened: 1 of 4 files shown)


rudigiesler:
Ready for review.

junebug/logging_service.py
```diff
@@ -79,7 +79,7 @@ def _log_to_file(self, event):
     def __call__(self, event):
         if self.log_context_sentinel in event:
             return
-        if self.worker_id not in event.get('system', '').split(','):
+        if self.worker_id not in (event.get('system', '') or '').split(','):
```
hodgestar: Maybe this should be `None` if we're doing the `or ''`?

hodgestar:
:+1:

junebug/logging_service.py
rudigiesler: Or we could just leave it out? Have `event.get('system')`

rudigiesler:
Added a small commit to simplify the `system` check for logs. Ready for another review.

hodgestar:
:+1:

PR #71, opened 2015-11-23
Create changelog in documentation

rudigiesler:
We need to add a changelog to the documentation to document what changed with each release of Junebug.

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -3,7 +3,7 @@
 Release Notes
 =============
 
-v1.0.0
+v0.1.0
 ------
 .. Pull requests 60,62-70
 
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -1,0 +1,105 @@
+.. _release-notes
+
+Release Notes
+=============
+
+v1.0.0
+------
+.. Pull requests 60,62-70
+
+- Features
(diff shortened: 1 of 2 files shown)


docs/release-notes.rst
```diff
@@ -0,0 +1,105 @@
+.. _release-notes
+
+Release Notes
+=============
+
+v1.0.0
```
rudigiesler: Should this be `v1.0.0`, or `v0.1.0`?

rudigiesler:
Ready for review.

docs/release-notes.rst
hodgestar: I think we're planning to do a splash `v1.0.0` release at some point in the future, so I'd vote for `v0.x.0` for some value of `x` less than 7 or so.

hodgestar:
Other than the version question, looks good.

docs/release-notes.rst
rudigiesler: Okay I'll do the `0.1.0` version then. I wanted something to signify that we've finished all the work that we originally set out to do, and not just continue with `0.0.6`.

rudigiesler:
Version changed, ready for review.

hodgestar:
:+1:

PR #76, opened 2015-12-10
Fix config file loading

rudigiesler:


rudigiesler:
This was a bug that I found when making the dockerfile. `yaml.safe_load` expects a file, not a filename. Ready for review.

rudigiesler:
Travis failure is because of new twisted version not supporting Python 2.6. I can't remember what we decided to do. Not support Python 2.6, or switch to an older version of twisted for 2.6?

hodgestar:
Looks good. Do we have a test for `load_config(None)`?

rudigiesler:
Test added for `load_config(None)`

hodgestar:
:+1:

PR #75, opened 2015-12-10
Create Dockerfile

rudigiesler:


rudigiesler:
Not yet ready for review. Relies on https://github.com/praekelt/junebug/pull/76 to merge and a release to be made before I can properly test this.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM debian:jessie
+MAINTAINER Praekelt Foundation <dev@praekeltfoundation.org>
+
+RUN apt-get -qq update && apt-get -qqy install python-pip python-dev
+RUN pip install junebug
+RUN mkdir logs
+
+EXPOSE 8080
+ENTRYPOINT ["jb", "--config=/app/example-config.yaml"]
```
hodgestar: I'm wondering if `example-config` is the right name? Perhaps we should have `example/config.yaml`and just have `example` mounted at `app`?

hodgestar: Do we want to set the current working directory to `/app`?

hodgestar:
Left a few comments, but seems good so far.

Dockerfile
rudigiesler: The reason why I didn't want to set the current working directory to `/app` is that we're mapping the folder as a volume, which would mean that we need a `logs` folder, and I didn't want that folder to be created on the host.

rudigiesler: `example-app` sounds good, I'll change it to that.

hodgestar: How do people view/keep their logs if they aren't on a volume though? Maybe
Junebug should create the logging folder if it doesn't exist? Not sure.

rudigiesler: Hmm. I'm not sure. I've got it setup to store junebug's logs in the app folder, but the channel specific logs will be stored in the container. Maybe we should have Junebug create the logging folder if it doesn't exist.

rudigiesler: @hodgestar What do you think about the idea of for now, creating a `channel-logs` folder in `example-app`, with a file so that git keeps that folder, and having the channel logs go there. And then creating a card to have junebug create the folder if it doesn't exist?

hodgestar: Sounds good!

rudigiesler:
Ready for review.

hodgestar:
We seem to be missing the change that sets the channel log folder to `/app/channel-logs` in the YAML config file?

rudigiesler:
Forgot to commit that change. Committed now. Ready for review.

hodgestar:
:+1:

PR #77, opened 2015-12-15
Forward MO messages to amqp queue

rudigiesler:


rudigiesler:
Ready for review

junebug/workers.py
```diff
@@ -114,6 +128,13 @@ def _store_event(self, event):
 
     @inlineCallbacks
     def _forward_event(self, event):
+        '''Forward the event to the correct places.'''
+        yield self._forward_event_http(event)
+        if self.config.get('message_queue') is not None:
```
hodgestar: Is there a reason this check is here rather than inside `_forward_event_amqp` like it is for `_forward_event_http`?

docs/http_api.rst
```diff
@@ -40,6 +40,13 @@ Channels
        to. May be ``null`` if not desired. Not supported by every channel.
    :param str mo_url:
        URL to call on incoming messages (mobile originated) from this channel.
+       One or both of mo_url or amqp_queue must be specified.
```
hodgestar: Should we perhaps mention that if both are specified, MO messages are sent to both?

docs/http_api.rst
```diff
@@ -40,6 +40,13 @@ Channels
...
+   :param str amqp_queue:
+       AMQP queue to repost messages onto for mobile originated messages. One
+       or both of mo_url or amqp_queue must be specified. Sends inbound
+       (mobile originated) messages to '{amqp_queue}.inbound', outbound (mobile
+       terminated) messages from '{amqp_queue}.outbound', and events to
+       '{amqp_queue}.events'
```
hodgestar: Should we mention that events are sent to event URLs too if those are specified?

I think we also need to move the MT bit into its own sentence because it's a bit confusing for me at the moment.

hodgestar:
Just checking that this PR isn't meant to implement processing of the MT inbound messages over the RabbitMQ queue?

junebug/workers.py
rudigiesler: No there isn't, I'll move it into `_forward_event_amqp` for uniformity.

rudigiesler:
Changes made. Ready for another review.

justinvdm:
:+1:

PR #83, opened 2016-03-15
Add environment variable for reactor select

rudigiesler:


junebug/__init__.py
```diff
@@ -2,6 +2,28 @@
 Junebug is a system for managing text messaging transports via a RESTful HTTP
 interface.
 '''
+import os
+
+r = os.environ.get('REACTOR')
```
hodgestar: I would prefix this with something. Maybe `TWISTED_REACTOR`? Or `JUNEBUG_REACTOR`?

junebug/command_line.py
```diff
@@ -118,12 +119,13 @@ def logging_setup(filename):
 
     LOGGING_FORMAT = '%(asctime)s [%(name)s] %(levelname)s: %(message)s'
 
-    # Send Twisted Logs to python logger
-    log.PythonLoggingObserver().start()
+    if not os.environ.get('DISABLE_LOGGING'):
```
hodgestar: Some thought about prefixing this env variable.

rudigiesler:
@hodgestar I've prefixed the environment variables. Ready for review

junebug/__init__.py
```diff
@@ -2,6 +2,28 @@
...
+    kqreactor.install()
+elif r == "WFMO":
+    from twisted.internet import win32eventreactor
+    win32eventreactor.install()
+elif r == "IOCP":
+    from twisted.internet import iocpreactor
+    iocpreactor.install()
+elif r == "EPOLL":
+    from twisted.internet import epollreactor
+    epollreactor.install()
```
hodgestar: Not wild about the giant `elif`, but at least it's simple. Perhaps we should put it in a function and call the function (just so it's a bit packaged up?

hodgestar:
Left one last comment, but otherwise looks good. Maybe also worthwhile adding a comment to the code and documenting the environment variables?

junebug/__init__.py
rudigiesler: Should we maybe rather to a `.get` on a dictionary?

rudigiesler:
@hodgestar I've added the environment variables to the documentation. I've also slightly neatened the large elif.

junebug/__init__.py
```diff
@@ -2,6 +2,26 @@
...
+    from twisted.internet import pollreactor as r
+elif r == "KQUEUE":
+    from twisted.internet import kqreactor as r
+elif r == "WFMO":
+    from twisted.internet import win32eventreactor as r
+elif r == "IOCP":
+    from twisted.internet import iocpreactor as r
+elif r == "EPOLL":
+    from twisted.internet import epollreactor as r
+if r is not None:
```
hodgestar: If someone sets `JUNEBUG_REACTOR` to `silly-typo`, this will attempt to call `.install()` on a string, which won't be good. Perhaps we should add a final `eilf`:
```python
elif r is not None:
    raise RuntimeError("Unsupported JUNEBUG_REACTOR setting %r" % (r,))
```
Might also be worthwhile allowing explicitly setting the string `DEFAULT` to get the default reactor (Twisted has `twisted.internet.default.install` if that helps).

rudigiesler:
@hodgestar I've addressed your comments. Ready for review again

hodgestar:
:+1:

PR #84, opened 2016-03-18
Documentation updates.

grigi:
Hi

This pull request Fixes .event instead of .events for AMQP messages.
It also updated the release notes, and expands a bit on AMQP integration.

### DIFF
--- a/docs/amqp-integration.rst
+++ b/docs/amqp-integration.rst
@@ -7,13 +7,13 @@ Should you choose to use the AMQP queues to handle messaging with Junebug, when
 you specify the ``amqp_queue`` parameter when you configure the
 :ref:`channel <channels>`, the usage is as follows:
 
-* Inbound *(mobile originated)* messages are sent to the **key** of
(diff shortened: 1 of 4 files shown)


docs/amqp-integration.rst
```diff
@@ -0,0 +1,48 @@
+.. _amqp-integration:
+
+AMQP integration
+================
+
+Should you choose to use the AMQP queues to handle messaging with Junebug, when
+you specify the ``amqp_queue`` parameter when you configure the
+:ref:`channel <channels>`, the usage is as follows:
+
+* Inbound *(mobile originated)* messages are sent to the **key** of
```
rudigiesler: Perhaps we should call it the `routing key` instead of `key`?

rudigiesler:
Just that one comment from me, otherwise it looks good. Thanks for updating this :)

grigi:
@rudigiesler Thanks, I did your suggestion. It definitely avoids some ambiguity :-)

rudigiesler:
:+1:

PR #86, opened 2016-04-05
Reload nginx when the nginx plugin starts.

hodgestar:
When the nginx plugin starts, it writes the vhosts file, but doesn't ask nginx to reload, so the vhosts file has no effect.


hodgestar:
@justinvdm @rudigiesler Ready for review.

justinvdm:
:+1;

PR #81, opened 2016-04-08
Allow for channel restarts.

hodgestar:
Channels can be reconfigured by POST-ing to the channel resource. Depending on what values are changed, a restart may be the result.

It would be nice if a simple POST resulted in a restart to deal with transports that've disconnected. Currently the only way to achieve that is to restart all of Junebug with all configured channels.

### DIFF
--- a/docs/http_api.rst
+++ b/docs/http_api.rst
@@ -169,6 +169,11 @@ Channels
    Delete a channel.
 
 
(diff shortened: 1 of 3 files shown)


hodgestar:
@rudigiesler @smn Ready for review.

smn:
:+1: :cake: thank you!

rudigiesler:
:+1:

PR #89, opened 2016-04-18
Add leading / to web_paths for nginx plugin.

hodgestar:
Channels that require a `web_path` don't need the `web_path` to have a leading `/` because the `web_path` is always considered to be relative to the root of the site. Nginx's location directives, however, do require a leading `/` since they're reading as a raw path prefix.

E.g. If `web_path` is `foo` we currently generate an Nginx config that looks like:
```
location foo { ... }
```
when we should be generating a config that looks like:
```
location /foo { ... }
```


hodgestar:
Ready for review.

smn:
:+1:

PR #90, opened 2016-04-25
API Errors don't say what key is the problem

rudigiesler:
When making a POST to change multiple fields on a channel that generates multiple errors the response does not indicate what fields caused the errors. For example when trying to set multiple fields to `null` the following response was generated:

```json
{
  "status": 400,
  "code": "Bad Request",
  "description": "api usage error",
  "result": {
    "errors": [
      {
        "message": "None is not of type 'string'",
        "type": "invalid_body"
      },
      {
        "message": "None is not of type 'string'",
        "type": "invalid_body"
      }
    ]
  }
}
```


rudigiesler:
Ready for review.

smn:
:+1:

PR #82, opened 2016-04-25
Benchmarking

rudigiesler:


rudigiesler:
Ready for review (maybe?)

smn:
What's the plan with this?

rudigiesler:
I think we want to be able to merge this into the main branch. I'm not too sure about including the benchmark results in the repo, but an example of using the ipython notebook to plot results might be useful, although I think it might be out of date, it was a while since last working on this.

PR #92, opened 2016-05-03
Fix MessageForwardingWorker teardown

rudigiesler:
Currently, the teardown_application method of the MessageForwardingWorker assumes that it has a redis manager, which won't be the case if the application didn't start up properly. We should add a safeguard against this.


rudigiesler:
Ready for review.

hodgestar:
:+1:

PR #93, opened 2016-05-03
Better handling for 301 responses

rudigiesler:
A 301 response currently looks like:

```{"status": 301, "code": "Moved Permanently", "description": null, "result": {"errors": [{"message": "301: Moved Permanently", "type": "Moved Permanently"}]}}```

We should at minimum include the URL that we are redirecting to.

Also, we should check that the headers are also being set correctly.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -82,11 +82,17 @@ class JunebugApi(object):
 
     @app.handle_errors(HTTPException)
     def http_error(self, request, failure):
+        error = {
+            'code': failure.value.code,
+            'type': failure.value.name,
+            'message': failure.getErrorMessage(),
+        }
+        if getattr(failure.value, 'new_url', None) is not None:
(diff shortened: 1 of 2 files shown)


rudigiesler:
Requirements given to us:
```We have set up the SMPP 2 SMS on the below account with the short code 36187:
 
Username: praekelt_rb
Password: AaTmob1l
 
You will need to bind to over the SMPP protocol as a Transmitter or a Transceiver with the following data:
Your Username: praekelt_rb
Your Password: AaTmob1l
 
Your Source Address : *03199 (Used to Submit and required by our system.)
Your Address Range : *03199 (Used to Bind and required by our system.)
 
Note: Your Source Address and Address Range will match.
 
This DNS : smsgw14.gsm.co.za
This Port : 2775```

rudigiesler:
Ready for review.

hodgestar:
:+1:

PR #100, opened 2016-05-30
Add libssl-dev and libffi-dev to install instructions

hodgestar:
Installing Junebug requires libssl-dev and libffi-dev to be present on Debian systems, so our install documents should mention those in the list of packages to apt-get install.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -3,11 +3,20 @@
 Installation
 ============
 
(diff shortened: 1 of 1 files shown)


hodgestar:
@ProfMerlin @rudigiesler Ready for review.

rudigiesler:
:+1:

PR #103, opened 2016-06-14
Specify minimum twisted version

rudigiesler:
The minimum twisted version that `vumi` has is `13.2.0`, which is too low, since we rely on new style classes in Junebug. We set a minimum version (`15.0`), and also run our travis tests against our minimum version.


hodgestar:
:+1:

PR #105, opened 2016-06-14
Modified logging_service.py in lines 115 and 116 to create logs direc…

munsanje:
…tory if it does not already exist @rudigiesler

### DIFF
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -31,10 +31,7 @@ class JunebugLogObserver(object):
         :type logfile: :class:`twisted.python.logfile.LogFile`
         :param str worker_id: ID of the worker that the log is for.
         '''
-<<<<<<< HEAD
 
-=======
->>>>>>> parent of 794a9f7... Modified logging_service.py in lines 115 and 116 to create logs directory if it does not already exist
         if log_context_sentinel is None:
             log_context_sentinel = DEFAULT_LOG_CONTEXT_SENTINEL
         self.worker_id = worker_id
@@ -115,12 +112,9 @@ class JunebugLoggerService(Service):
         self.max_files = max_files
 
     def startService(self):
-<<<<<<< HEAD
 
         if not os.path.exists(os.path.dirname(self.path)):
-            os.makedirs(os.path.dirname(self.path),0777)
-=======
->>>>>>> parent of 794a9f7... Modified logging_service.py in lines 115 and 116 to create logs directory if it does not already exist
+            os.makedirs(self.path,0777)
         self.logfile = LogFile(
             self.worker_id, self.path, rotateLength=self.rotate,
             maxRotatedFiles=self.max_files)
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -31,6 +31,7 @@ class JunebugLogObserver(object):
         :type logfile: :class:`twisted.python.logfile.LogFile`
         :param str worker_id: ID of the worker that the log is for.
         '''
+        #os.makedirs("log", 0777)
         if log_context_sentinel is None:
             log_context_sentinel = DEFAULT_LOG_CONTEXT_SENTINEL
         self.worker_id = worker_id
@@ -111,6 +112,8 @@ class JunebugLoggerService(Service):
         self.max_files = max_files
 
     def startService(self):
+        if not os.path.exists(self.path):
+            os.makedirs(self.path,0777)
         self.logfile = LogFile(
             self.worker_id, self.path, rotateLength=self.rotate,
             maxRotatedFiles=self.max_files)
--- a/junebug/logging_service.py
+++ b/junebug/logging_service.py
@@ -31,7 +31,6 @@ class JunebugLogObserver(object):
         :type logfile: :class:`twisted.python.logfile.LogFile`
         :param str worker_id: ID of the worker that the log is for.
         '''
-
         if log_context_sentinel is None:
             log_context_sentinel = DEFAULT_LOG_CONTEXT_SENTINEL
         self.worker_id = worker_id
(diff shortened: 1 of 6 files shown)


junebug/logging_service.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, logfile, worker_id, log_context_sentinel=None):
         :type logfile: :class:`twisted.python.logfile.LogFile`
         :param str worker_id: ID of the worker that the log is for.
         '''
+
```
hodgestar: We prefer not to have newlines here, and more generally it's polite to not unnecessarily modify code that isn't directly related to the changes needed for the new feature (although lines can be blurry).

junebug/service.py
```diff
@@ -6,7 +6,6 @@
 
 from junebug import JunebugApi
 
-
```
hodgestar: Since we're no longer changing this file we should undo this line removal.

junebug/logging_service.py
```diff
@@ -111,6 +112,9 @@ def __init__(self, worker_id, path, rotate, max_files, logger=None):
         self.max_files = max_files
 
     def startService(self):
+
```
hodgestar: I would remove the blank line here to match how we usually do this.

junebug/logging_service.py
```diff
@@ -111,6 +112,9 @@ def __init__(self, worker_id, path, rotate, max_files, logger=None):
...
+        if not os.path.exists(os.path.dirname(self.path)):
+            os.makedirs(os.path.dirname(self.path),0777)
```
hodgestar: PEP8 likes a space after a `,`.

GitHub seems to have highlighted the `0777` in red. I'm not entirely sure why. One can also use `0o777` for octal literals though, which is perhaps a little more explicit and maybe GitHub will not highlight in red. Up to you. :)

junebug/tests/test_logging_service.py
```diff
@@ -207,6 +210,20 @@ def test_start_stop(self):
...
+            os.makedirs(self.service.path, 0777)
+            self.stat1 = os.stat(self.service.path)
+            yield self.service.startService()
+            self.stat2 = os.stat(self.service.path)
+            self.assertEqual(self.stat1, self.stat2)
+            yield self.service.stopService()
+        yield self.service.startService()
+        self.assertTrue(os.path.exists(self.service.path))
+        yield self.service.stopService()
+
```
rudigiesler: We probably also want a test to make sure that if the directory doesn't exist, that it is actually created when we start the service.

junebug/tests/test_logging_service.py
```diff
@@ -207,6 +210,20 @@ def test_start_stop(self):
...
+        if not os.path.exists(self.service.path):
+            os.makedirs(self.service.path, 0777)
+            self.stat1 = os.stat(self.service.path)
+            yield self.service.startService()
+            self.stat2 = os.stat(self.service.path)
+            self.assertEqual(self.stat1, self.stat2)
+            yield self.service.stopService()
+        yield self.service.startService()
+        self.assertTrue(os.path.exists(self.service.path))
+        yield self.service.stopService()
```
rudigiesler: I'm not quite sure why lines 223-225 are needed, what is their purpose?

junebug/tests/test_logging_service.py
```diff
@@ -207,6 +210,20 @@ def test_start_stop(self):
...
+    @inlineCallbacks
+    def test_dir_create(self):
+        '''If log directory already exists, make sure it is not recreated'''
+        if not os.path.exists(self.service.path):
+            os.makedirs(self.service.path, 0777)
+            self.stat1 = os.stat(self.service.path)
+            yield self.service.startService()
+            self.stat2 = os.stat(self.service.path)
+            self.assertEqual(self.stat1, self.stat2)
+            yield self.service.stopService()
```
rudigiesler: I think that lines 218-222 should not be part of the if statement. If the path doesn't exist, we should create it, that's fine. But we should not only run our checks if the path doesn't exist. If something changes in the future that causes that path to be created before this test runs, that means that none of these checks will be executed.

rudigiesler:
Just a few comments from me, looks good otherwise :)

junebug/tests/test_logging_service.py
munsanje: I wrote lines 218-222 to check that if a directory already exists, it isn't overwritten by startService. The comparison in line 221 should compare the old and new metadata to check this.

munsanje: I should probably take them out of the if statement though.

junebug/tests/test_logging_service.py
munsanje: They are meant to test that the log directory exists after `startService()` is run. I've changed it to remove any existing log directory, run `startService()`, and then check that a directory exists.

junebug/tests/test_logging_service.py
```diff
@@ -207,6 +211,22 @@ def test_start_stop(self):
         self.assertFalse(self.service.registered())
         self.assertEqual(self.service.logfile.closed_count, 1)
 
+    @inlineCallbacks
+    def test_dir_create(self):
+        '''If log directory already exists, make sure it is not recreated.'''
+        if not os.path.exists(self.service.path):
+            os.makedirs(self.service.path, 0777)
+        self.stat1 = os.stat(self.service.path)
```
hodgestar: We can just use a local variable here since there's no need to access `stat1` outside of this method. (so `stat1 = ...` rather than `self.stat1 = ...`). The same goes for `stat2`.

junebug/tests/helpers.py
```diff
@@ -189,6 +192,7 @@ def create_channel(
 
         properties = deepcopy(properties)
         logpath = self.mktemp()
+        os.mkdir(logpath)
```
hodgestar: Do we need these `os.mkdir(logpath)` statements? It seems a bit odd that they're needed now when we're changing the code to create the log folders if they don't exist?

rudigiesler:
Ready for review.

junebug/logging_service.py
```diff
@@ -111,6 +111,8 @@ def __init__(self, worker_id, path, rotate, max_files, logger=None):
         self.max_files = max_files
 
     def startService(self):
+        if not os.path.exists(self.path):
+            os.makedirs(self.path, 0777)
```
smn: Do we want 755 instead? Only read + exec for everyone, read, exec & write for the owner?

smn:
:+1: other than the relatively minor question about permissions

rudigiesler:
@smn Changes made. Ready for another review.

smn:
:+1:

PR #108, opened 2016-09-15
Don't store events without user_message_id

rudigiesler:
Currently, if Junebug gets sent an event by the transport, but the `user_message_id` of the event is `null`, the event processing crashes and gets stuck. We should just log that the event doesn't have a message id, don't store it, and carry on.


rudigiesler:
Ready for review.

smn:
:+1:

PR #112, opened 2017-01-18
Change SmppTransport to SmppTransceiverTransport

rudigiesler:
In vumi, the SmppTransport is actually the depreciated SmppTransceiverTransportWithOldConfig. We should change this over to the new SmppTransceiverTransport.


rudigiesler:
Ready for review

PR #91, opened 2017-02-01
Updating channels to unset status_url, etc.

smn:
Currently one isn't allowed to post values for status_url that are `null` when updating a channel, which means there is no way to unset a status URL. We probably need to relax the validation on status_url and a few other fields to allow them to be null in addition to strings, so, e.g.:

``` python
"status_url": {"type": ["string", "null"]},
```

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -175,7 +175,7 @@ class JunebugApi(object):
                 'config': {'type': 'object'},
                 'metadata': {'type': 'object'},
                 'status_url': {'type': 'string'},
-                'mo_url': {'type': 'string'},
+                'mo_url': {'type': ['string', 'null']},
                 'rate_limit_count': {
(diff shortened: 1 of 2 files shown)


smn:
@rudigiesler ready for review

junebug/api.py
```diff
@@ -175,7 +175,7 @@ def get_channel(self, request, channel_id):
                 'config': {'type': 'object'},
                 'metadata': {'type': 'object'},
                 'status_url': {'type': 'string'},
-                'mo_url': {'type': 'string'},
+                'mo_url': {'type': ['string', 'null']},
```
rudigiesler: I think we want to do `status_url` as well, I think it's the only other one that we could accept `null` on.

smn:
@rudigiesler ready for review agian

PR #114, opened 2017-02-01
Add timeout for HTTP requests

smn:
Currently, we will wait forever for an outbound HTTP request (from us to another service) to return. We should add a configurable timeout, with a sensible default value. `treq` provides a `timeout` parameter for us to do this.

### DIFF
--- a/junebug/workers.py
+++ b/junebug/workers.py
@@ -23,6 +23,16 @@ class MessageForwardingConfig(ApplicationConfig):
         "The URL to send HTTP POST requests to for MO messages",
         default=None, static=True)
 
+    mo_message_url_timeout = ConfigInt(
+        "Maximum time (in seconds) a mo_message_url is allowed to take "
+        "to process a message",
+        default=10, static=True)
+
+    event_url_timeout = ConfigInt(
+        "Maximum time (in seconds) an event_url is allowed to take "
+        "to process an event",
+        default=10, static=True)
+
     message_queue = ConfigText(
         "The AMQP queue to forward messages on",
         default=None, static=True)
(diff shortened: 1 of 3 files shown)


smn:
@rudigiesler ready for review

junebug/workers.py
```diff
@@ -242,8 +262,9 @@ def request_failed(resp):
...
 
 
-def post(url, data):
+def post(url, data, timeout):
     return treq.post(
         url.encode('utf-8'),
         data=json.dumps(data, cls=JSONMessageEncoder),
-        headers={'Content-Type': 'application/json'})
+        headers={'Content-Type': 'application/json'},
+        timeout=timeout)
```
rudigiesler: If the timeout is reached, this would raise an exception. I think we should probably catch that exception and log it, rather than letting it rise up.

There are a few exceptions to catch with this: https://github.com/twisted/treq/issues/117

smn: Shall I catch these here in `post()` or a level up where `post()` is called?

rudigiesler: I think in `post()` is a good place; we need to do the same thing for both messages and events (catch and log the error), so we would just be repeating ourselves.

smn:
@rudigiesler ready for rereview

junebug/workers.py
```diff
@@ -86,8 +99,10 @@ def consume_user_message(self, message):
         msg = api_from_message(message)
 
         if self.config.get('mo_message_url') is not None:
-            resp = yield post(self.config['mo_message_url'], msg)
-            if request_failed(resp):
+            config = self.get_static_config()
+            resp = yield post(config.mo_message_url, msg,
+                              timeout=config.mo_message_url_timeout)
+            if resp and request_failed(resp):
```
smn: because we're handling the errors in `post()` the response here can be none.

PR #119, opened 2017-03-30
Messages API endpoint documentation is incomplete

smn:
The messages API endpoint states that both `reply_to` and `to` cannot be specified at the same time, but it does not specify that also `reply_to` and `from` may not be specified at the same time.

The docs should be updated to reflect that.


smn:
@rudigiesler ready for review

smn:
these build failures look entirely unrelated to these doc changes o_O

smn:
@rudigiesler build failure fixed

Issue #99, opened 2016-05-27
Add support for Basic Authentication in outbound HTTP requests


hodgestar:
Currently Junebug supports supplying URLs for sending messages, message status updates and channel status updates to. These URLs should allow a username and password to be specified within the URL and have those converted into Basic Authentication HTTP headers when HTTP requests are made to them.

E..g `https://user:pw@example.com/` should become a request to `https://example.com/` with `user` and `pw` supplied as the basic authentication username and password.

imsickofmaps:
@rudigiesler can you point me at the code that needs fixing for this? Is it in Junebug (I couldn't see where mo_url is consumed in this repo) or deeper in Vumi?

When we try to use Basic Auth in mo_url we get stuff like the following in the logs:
 ```
File "/usr/local/lib/python2.7/site-packages/idna/core.py", line 253, in check_label
5/17/2017 3:59:57 PM    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
5/17/2017 3:59:57 PMidna.core.InvalidCodepoint: Codepoint U+003A at position 8 of u'junebug:password@api' not allowed
```

rudigiesler:
@imsickofmaps The code that makes the http request is here: https://github.com/praekelt/junebug/blob/develop/junebug/workers.py#L282-L289 . We use `treq` for the request, and I'm guessing that treq doesn't support basic auth in the URL, or maybe we need to set a flag or something, we would investigate treq to see what can be done there.

If treq doesn't support it, then we need to extract the auth details from the URL, and pass the auth details and URL as separate fields to treq.

imsickofmaps:
@rudigiesler properly supporting auth seems like a better solution for Junebug. Treq does support basic auth, as a param on the request: http://treq.readthedocs.io/en/latest/api.html#treq.request

rudigiesler:
@imsickofmaps In that case, we'll need to extract the basic auth details from the URL, and pass the URL and auth details as separate parameters to treq.

smn:
Implemented in #129

PR #125, opened 2017-05-19
Issue 99 basic auth support

imsickofmaps:
Unsure right way to test this with your helper setup... 

Resolves: https://github.com/praekelt/junebug/issues/99


imsickofmaps:
@rudigiesler note, no new tests... unsure how with your helper setup.

rudigiesler:
@imsickofmaps So the test should be similar to https://github.com/praekelt/junebug/blob/develop/junebug/tests/test_workers.py#L65-L77 , except instead of testing the body, you would test that the request headers are correctly set. 

To add a custom mo_url, you would start a new worker, with a custom config, similar to https://github.com/praekelt/junebug/blob/develop/junebug/tests/test_workers.py#L60-L63 , except you would be setting the `mo_message_url` field, instead of the `transport_name` field. You can then use `worker.consume_user_message`, to use the newly created `worker`, instead of `self.worker.consume_user_message`.

You can find the URL of the dummy receiving API at `self.url`, you'll have to split it, add the `username@password` bit, and joining it up again, before passing it in the config.

smn:
Implemented in #129

PR #126, opened 2017-06-06
Junebug sentry exception handling

erikh360:


junebug/command_line.py
```diff
@@ -118,6 +118,13 @@ def parse_arguments(args):
     return config_from_args(vars(parser.parse_args(args)))
 
 
+class PythonInfoLoggingObserver(log.PythonLoggingObserver):
```
rudigiesler: I'm not sure that this is the best name for this.. Maybe `PythonExceptionFilteringLoggingObserver`, and a docstring to describe that it filters out exceptions, since they go to sentry.

Also, what happens with this when we don't select sentry logging? We still want these things logged when we don't have the sentry listener.

PR #127, opened 2017-06-09
Installing Junebug on py3 fails because of tiny setup.py bug

smn:
The description is a tuple, for some reason this is not a problem in 
py2 but in py3 it is.


smn:
@rudigiesler @miltontony @erikh360 tiny bug fix

PR #129, opened 2017-07-10
Allow auth on mo_url

smn:
Auth can either be basic auth:

```mo_url: 'https://foo:bar@example.org/'```

or via token auth:

```mo_url_auth_token: 'the token to use'```


smn:
@miltontony @KaitCrawford @jpaidoussi ready for review

smn:
This is a backwards compatible change.

mracter:
+1

PR #130, opened 2017-07-14
Add newline so reStructuredText renders correctly

alexmuller:
https://junebug.readthedocs.io/en/latest/cli-reference.html


smn:
:+1:

PR #132, opened 2017-07-17
Add basic and token auth as options for event urls

KaitCrawford:


junebug/stores.py
```diff
@@ -89,7 +89,7 @@ def load_vumi_message(self, channel_id, message_id):
 class OutboundMessageStore(BaseStore):
     '''Stores the event url, in order to look it up when deciding where events
     should go'''
-    PROPERTY_KEYS = ['event_url']
+    PROPERTY_KEYS = ['event_url', 'event_auth_token']
```
smn: For consistency with `mo_url_auth_token` I would suggest we use `event_url_auth_token` even though it is quite verbose.

junebug/workers.py
```diff
@@ -201,14 +201,47 @@ def _forward_event_http(self, event):
...
+        auth_token = yield self._get_event_auth_token(event)
+
+        if auth_token:
+            headers = {
+                'Authorization': [
+                    'Token %s' % (auth_token,)]
+            }
+        else:
+            headers = {}
+
```
smn: Do we perhaps want to remove the duplication of parsing auth creds from the url here? Perhaps a shared function for both mo & events since they're largely doing the same thing here.

smn:
🍰 Nice work @KaitCrawford sorry I missed the ping on this earlier. Just minor comment about the name of the key & possibly being able to reduce code duplication when parsing urls.

KaitCrawford:
@dubesoftware here's the code we worked on. You'll note I changed it so that we're storing the token rather than the whole auth dict. While writing the tests I come across some complications and concerns that suggested this was a better way to do it

KaitCrawford:
@smn this is ready for re-review thanks

smn:
:+1:

smn:
thank you!

PR #133, opened 2017-07-18
allow any 6.x.x version

smn:
Raven 6.x.x was released in February, allow it to be used.


smn:
@alexmuller @erikh360 @KaitCrawford ready for review

PR #141, opened 2017-11-21
Health endpoint for stuck messages

erikh360:
This endpoint calls the rabbitmq api to investigate the queues and make sure the messages are being cleared.

It will only work for rabbitmq.


junebug/api.py
```diff
@@ -301,3 +304,33 @@ def get_message_status(self, request, channel_id, message_id):
     @app.route('/health', methods=['GET'])
     def health_status(self, request):
         return response(request, 'health ok', {})
+
+    @app.route('/channels_health', methods=['GET'])
+    @inlineCallbacks
+    def channels_health(self, request):
```
rudigiesler: We should probably put this in with the normal channel health, rather than having a separate endpoint for it.

This should also be behind a config flag that is by default disabled, and have documentation around if you're running rabbitmq, then you can enable this to get this extra functionality

junebug/api.py
```diff
@@ -301,3 +304,33 @@ def get_message_status(self, request, channel_id, message_id):
...
+        channels_stuck = []
+
+        base_url = "http://%s:%s/api/queues/%s" % (
+            self.amqp_config['hostname'],
+            self.amqp_config.get('api_port', '15672'),
```
rudigiesler: So by default on rabbitmq, `5672` is the main port, and `15672` is the management port. So if I'm specifying my rabbitmq, I would usually specify a port of `5672`, and then this code will try to get management things from the incorrect port.

We will need a separate config value for the rabbitmq management interface port.

junebug/api.py
```diff
@@ -301,3 +304,33 @@ def get_message_status(self, request, channel_id, message_id):
...
+            self.amqp_config['vhost'].replace('/', '%2F'))
```
rudigiesler: We should rather urlencode the entire vhost string, rather than just urlencoding the `/` character, otherwise there might be other characters in the vhost name that we might run into that need to be encoded.

junebug/api.py
erikh360: `api_port` is a new config value for when your management is on something other than `15672`. `port` is the existing one. 

I'll document appropriately so it makes more sense.

rudigiesler: Maybe we should then name it `management_port` or something like that, so it's more obvious?

junebug/api.py
```diff
@@ -299,5 +303,35 @@ def get_message_status(self, request, channel_id, message_id):
...
+
+            ids = yield Channel.get_all(self.redis)
+
+            for channel_id in ids:
+                for sub in ['inbound', 'outbound', 'event']:
+                    url = "%s/%s.%s" % (base_url, channel_id, sub)
+
+                    queue = requests.get(url, auth=requests.auth.HTTPBasicAuth(
+                        self.amqp_config['username'],
+                        self.amqp_config['password'])).json()
```
rudigiesler: Maybe we should have a client class for using the rabbitmq management interface? There are some existing clients that we can evaluate to use, to avoid having to make our own.

junebug/tests/helpers.py
```diff
@@ -154,7 +154,14 @@ class JunebugTestBase(TestCase):
...
         'ttl': 60,
-        'amqp': {},
+        'amqp': {
+            'hostname': 'rabbitmq',
+            'vhost': '/',
+            'port': '5672',
+            'db': '0',
+            'username': 'guest',
+            'password': 'guest'
+        },
```
rudigiesler: I don't think we need to specify all of these parameters, since they already have defaults in the configuration file

https://github.com/praekelt/junebug/blob/e8e3e51710a578009cbebdb05048af4380bb5919/junebug/config.py#L32-L41

junebug/api.py
```diff
@@ -299,5 +301,44 @@ def get_message_status(self, request, channel_id, message_id):
...
+
+            queues = []
+            stuck = False
+
+            for channel_id in ids:
+                for sub in ['inbound', 'outbound', 'event']:
+                    queue_name = "%s.%s" % (channel_id, sub)
+
+                    queue = rabbit_client.get_queue(
+                        self.amqp_config['vhost'], queue_name)
```
rudigiesler: This is doing a blocking request, which is very bad in twisted land. If there's no way to have this client do a non-blocking request, or if there aren't any non blocking clients, then we might have to create our own (we probably only need to implement this one endpoint for now though)

junebug/command_line.py
```diff
@@ -114,6 +114,11 @@ def create_parser():
         '--max-logs', '-ml', type=int,
         dest='max_logs', help='the maximum number of log entries to '
         'to allow to be fetched through the API. Defaults to 100.')
+    parser.add_argument(
+        '--rabbitmq-management-interface', '-rmi',
+        dest='rabbitmq_management_interface', type=str,
+        help='If `True`, the health of each individual channel will be '
+        'checked. This is only available for RabbitMQ')
```
rudigiesler: I think this help text is incorrect.

junebug/config.py
```diff
@@ -85,3 +85,8 @@ class JunebugConfig(Config):
     max_logs = ConfigInt(
         "The maximum amount of logs that is allowed to be retrieved via the "
         "API.", default=100)
+
+    rabbitmq_management_interface = ConfigText(
+        "If set, the health of each individual queue will be checked. "
+        "This is only available for RabbitMQ",
```
rudigiesler: Same with this help text. This should be the url string of the rabbitmq management interface.

junebug/api.py
```diff
@@ -338,5 +356,70 @@ def create_router(self, request, body):
...
+                            queue_name
+                        )
+                        http_client = HTTPClient(
+                            self.agent_factory(
+                                self.clock,
+                                pool=self.pool_factory(self.clock)))
+
+                        get = http_client.get(
+                            url, auth=(self.amqp_config['username'],
+                                       self.amqp_config['password']))
```
rudigiesler: This is all quite messy having these things nested so deep. Could we put all the making the url and HTTP request things into a rabbitmq management client class, so that you just go `rabbitmq_management_client.get_queue(vhost, queue_name)`

junebug/api.py
```diff
@@ -338,5 +356,70 @@ def create_router(self, request, body):
...
+
+            def get_queues(channel_ids):
+                gets = []
+                for channel_id in channel_ids:
+                    for sub in ['inbound', 'outbound', 'event']:
+
+                        queue_name = "%s.%s" % (channel_id, sub)
+                        url = 'http://%s/api/queues/%s/%s' % (
+                            self.config.rabbitmq_management_interface,
+                            self.amqp_config['vhost'].replace('/', '%2F'),
```
rudigiesler: I think we want to urlencode all characters in the vhost, not just the `/`

junebug/api.py
```diff
@@ -338,5 +356,70 @@ def create_router(self, request, body):
         ))
 
     @app.route('/health', methods=['GET'])
+    @inlineCallbacks
     def health_status(self, request):
-        return response(request, 'health ok', {})
+        if self.config.rabbitmq_management_interface:
+
+            ids = yield Channel.get_all(self.redis)
```
rudigiesler: We can probably do away with the `inlineCallbacks` if we chain the `get_queues` function to this one, ie.
```
d = Channel.get_all(self.redis)
d.addCallback(get_queues)
d.addCallback(defer.DeferredList)
d.addCallback(return_queue_results)
return d
```

junebug/api.py
```diff
@@ -338,5 +356,70 @@ def create_router(self, request, body):
...
+                code = http.OK
+                if stuck:
+                    status = "channels stuck"
+                    code = http.INTERNAL_SERVER_ERROR
+
+                return response(request, status, queues, code=code)
+
+            outputs = get_queues(ids)
+
+            final = defer.DeferredList(outputs)
```
rudigiesler: The way we're doing it here, we're making all the requests in parallel. The issue with this is that firstly, the connection pool will only be used for multiple requests to `/health`, so I don't think that it's doing much. 

Secondly, if the amount of requests is greater than the pool size, then those extra connections won't be cached.

Thirdly, I don't think that the management interface will appreciate it much if we're making hundreds of parallel requests to it, at some amount of parallel requests I'm sure that the requests will start failing.

I think we need to limit the amount of requests that we're making at any time, I think we can use twisted's DeferredSemaphore to do this, some nice explanations and examples here: https://patrickmn.com/software/using-twisteds-deferred-semaphore/

junebug/tests/test_api.py
```diff
@@ -25,6 +29,10 @@ def get(self, url, params={}):
         return treq.get(
             "%s%s" % (self.url, url), params=params, persistent=False)
 
+    def request(self, method, url, params={}):
+        return treq.request(
+            method, "%s%s" % (self.url, url), params=params, persistent=False)
```
rudigiesler: I'm not sure that we need this function, we only seem to be using the `GET` method, and that functionality is fully covered by the `get` function above.

junebug/api.py
erikh360: Ah, missed this in my cleanup.

junebug/tests/test_api.py
rudigiesler: Ah okay, I see 👍

junebug/api.py
```diff
@@ -25,15 +41,46 @@ class ApiUsageError(JunebugError):
...
         self.config = config
 
+    @classmethod
+    def pool_factory(cls, reactor):
+        pool = HTTPConnectionPool(reactor, persistent=True)
+        pool.maxPersistentPerHost = TPS_LIMIT
+
+    @classmethod
+    def agent_factory(cls, reactor, pool=None):
+        return Agent(reactor, pool=pool)
```
rudigiesler: I think that the client can have its own pool and agent, and manage its own connection pool.

docs/http_api.rst
```diff
@@ -752,3 +752,16 @@ A status event can be categorised under one of the following levels:
...
+Health
+^^^^^^
+
+.. http:get:: /health/
+
+   Provides HTTP GET access to test or verify the health of the system.
+
+   If the ``rabbitmq_management_interface`` config item is set it will also
+   query the RabbitMQ Management interface to check the health of each queue.
+   This is only available for RabbitMQ.
```
rudigiesler: I think the important thing here is just to add that a 200 is returned if the health is okay, and a 500 is returned if it's not. Maybe also an example of the JSON returned if the `rabbitmq_management_interface` config item is set, so that people can see the format of the response in case they want the details.

docs/release-notes.rst
```diff
@@ -29,6 +29,13 @@ v0.1.27 (Not yet released)
     - Update Readme to give more information to someone who wants to work on
       the project.
 
+.. Pull requests 141
+*24 November 2017*
```
rudigiesler: I think we can just update the date and pull request list above, since none of this has been released yet. I only put that date above as a temporary thing to be replaced when we do the actual release.

junebug/api.py
```diff
@@ -339,4 +343,58 @@ def create_router(self, request, body):
 
     @app.route('/health', methods=['GET'])
     def health_status(self, request):
-        return response(request, 'health ok', {})
+        if self.config.rabbitmq_management_interface:
+            rabbitmq_management_client = RabbitmqManagementClient(
+                self.config.rabbitmq_management_interface,
+                self.amqp_config['username'],
+                self.amqp_config['password'])
```
rudigiesler: This will be creating a new client for every request, which could mean a lot of parallel requests to the management interface if Junebug is handling a lot of parallel requests to this endpoint.

This is probably a rare case, but maybe we can create the client in `setup`, if the config value is set.

PR #142, opened 2017-11-22
Add list and create endpoints for routers

rudigiesler:


dubesoftware:
Looks about right and is syntactically sound. My current knowledge of Junebug internals doesn't allow for more analysis.

junebug/router/base.py
```diff
@@ -0,0 +1,109 @@
...
+        if self.junebug_config.replace_routers:
+            return self.junebug_config.routers
+        else:
+            routers = {}
+            routers.update(default_router_types)
+            routers.update(self.junebug_config.routers)
+            return routers
+
+    @property
+    def _worker_class_name(self):
```
KaitCrawford: I don't see a test for this. Am I missing it or are you planning to add it later?

rudigiesler: Ah, no it seems there isn't. I'll add it as a test for the `start` function.

rudigiesler:
@KaitCrawford Changes addressed, ready for another review

PR #144, opened 2017-11-24
Add updating (PUT and PATCH) router endpoints

rudigiesler:


rudigiesler:
@KaitCrawford When you first restore a router object from an ID, it gets the information from Redis. From then on, it uses the information stored on the router object. You can update the information stored in Redis at any time by calling the `.save()` function.

On the PUT and PATCH endpoints, you'll see that we restore the router object, which will fetch the information from Redis. Then we modify the config, validate the config (the validation uses whatever is stored on the class object, it doesn't fetch from Redis again), stop and start the router (the start function also uses what is stored on the class object, it doesn't fetch from Redis again), and then we call save, once we know that everything is running, so that if anything is broken, at least those broken changes don't get saved into Redis.

PR #148, opened 2017-11-30
Add list and get endpoints for router destinations

rudigiesler:


junebug/api.py
```diff
@@ -461,6 +461,28 @@ def create_router_destination(self, request, body, router_id):
...
+
+    @app.route(
+        '/routers/<string:router_id>/destinations/<string:destination_id>',
+        methods=['GET'])
+    @inlineCallbacks
+    def get_destination(self, request, router_id, destination_id):
+        """Get the config and status of a destination"""
+        router = yield Router.from_id(
+            self.router_store, self.config, self.service, router_id)
+        destination = router.get_destination(destination_id)
```
erikh360: Shouldn't this function be called with a `yield` or a `addCallBack`?

rudigiesler: Nope, that function doesn't return a deferred. The destination is loaded from config in the `from_id` function, so it's just returning what it has in memory.

It's loaded in the beginning because destinations aren't really separate things, they're just a list in the router worker's config, so it makes sense to load all of the destinations when we restore the router.

erikh360: Makes sense.

PR #150, opened 2017-12-05
Add delete endpoint for router destinations

rudigiesler:


junebug/tests/test_api.py
```diff
@@ -1727,3 +1727,26 @@ def test_update_destination_config_non_existing_destination(self):
                     'type': 'DestinationNotFound',
                 }],
             })
+
+    @inlineCallbacks
+    def test_delete_destination(self):
+        """If the destination doesn't exist, then a not found error should be
```
erikh360: This doesn't match the test, but would be a nice extra test.

rudigiesler:
@erikh360 Comments addressed :) ready for another review

PR #157, opened 2017-12-13
Fix mo_url_auth_token in docs

alexmuller:
`mo_url_token` doesn't do anything, I think it should be `mo_url_auth_token`.

Not sure if this applies both to channels and routers?


rudigiesler:
Router destinations are not the same :) https://github.com/praekelt/junebug/blob/develop/junebug/api.py#L440 . I built those endpoints from the docs, and I built the docs from the existing docs, so now they take different parameters :/

alexmuller:
Ah right, updated!

PR #163, opened 2018-01-10
Create log endpoint for routers

erikh360:


junebug/tests/helpers.py
```diff
@@ -284,6 +288,18 @@ def create_channel_from_id(self, redis, config, id, service):
         channel = yield Channel.from_id(redis, config, id, service)
         returnValue(channel)
 
+    @inlineCallbacks
+    def create_logging_router(self, service, config={}):
```
rudigiesler: We're not really creating a router specific to logging here, this is just a general test router. So maybe a better name for this function is `create_test_router`?

junebug/tests/test_api.py
```diff
@@ -1460,6 +1460,141 @@ def test_delete_non_existing_router(self):
...
+    @inlineCallbacks
+    def test_get_router_logs_more_than_configured(self):
+        '''If the amount of requested logs is more than what is
+        configured, then only the configured amount of logs are returned.'''
+        logpath = self.mktemp()
+        config = yield self.create_channel_config(
+            max_logs=2,
+            channels={
+                'logging': 'junebug.tests.helpers.LoggingTestTransport',
+            },
```
rudigiesler: I don't think that we need this channels bit, since we're not creating or using any channels.

junebug/tests/test_api.py
```diff
@@ -1460,6 +1460,141 @@ def test_delete_non_existing_router(self):
...
+    @inlineCallbacks
+    def test_get_router_logs_no_n(self):
+        '''If the number of logs is not specified, then the API should return
+        the configured maximum number of logs.'''
+        logpath = self.mktemp()
+        config = yield self.create_channel_config(
+            max_logs=2,
+            channels={
+                'logging': 'junebug.tests.helpers.LoggingTestTransport',
+            },
```
rudigiesler: Same here

rudigiesler (approved):

PR #164, opened 2018-01-11
Create message endpoints

erikh360:


junebug/api.py
```diff
@@ -584,6 +556,97 @@ def delete_router_destination(self, request, router_id, destination_id):
...
+            },
+            'required': ['content'],
+            'additionalProperties': False,
+        }))
+    @inlineCallbacks
+    def send_destination_message(
+            self, request, body, router_id, destination_id):
+        '''Send an outbound (mobile terminated) message'''
+        router = yield Router.from_id(self, router_id)
+        router.get_destination(destination_id)
```
rudigiesler, line 585: I'm not sure why we need to get the destination here?

junebug/api.py
```diff
@@ -584,6 +556,97 @@ def delete_router_destination(self, request, router_id, destination_id):
...
+        channel_id = router.router_worker.get_destination_channel(
+            destination_id)
```
rudigiesler: Maybe this should be an async method, in case further implementations need to do lookups in the message store, or something like that.

For the existing method, you can just do a `return succeed(value)` to make it return a deferred. Then, since you're using inlineCallbacks here, you can just `yield` this line.

junebug/api.py
```diff
@@ -584,6 +556,97 @@ def delete_router_destination(self, request, router_id, destination_id):
...
+    @app.route(
+        '/routers/<string:router_id>/destinations/<string:destination_id>/messages/<string:message_id>',  # noqa
+        methods=['GET'])
+    @inlineCallbacks
+    def get_destination_message_status(
+            self, request, router_id, destination_id, message_id):
+
+        router = yield Router.from_id(self, router_id)
+        channel_id = router.router_worker.get_destination_channel(
+            destination_id)
```
rudigiesler: Maybe we should also be passing the message body, or something about the message, into this method, in case future routers don't have a simple mapping of channel -> destination (eg we want to route depending on which MNO the user is on)

junebug/api.py
erikh360: This is to make sure the destination actually exists.

rudigiesler: We probably need to do this on the get message status endpoint then too, and maybe a test for that endpoint.

junebug/api.py
erikh360: I'm not sure if this is possible in the `GET`.

rudigiesler: The message should be in the message store, so we should be able to load the body from the message store, and pass it to the function?

erikh360: To get the message from the store I need the channel id.

rudigiesler: Ah right. Well maybe we can send a message ID on the `GET`, and the message body on the `POST`? That way, on the `GET`, it can check each channel's message store until it finds the message, and the POST it already has the message body, so it knows where to route.

erikh360:
@rudigiesler this is ready for another look, tomorrow.

junebug/router/base.py
```diff
@@ -356,6 +356,12 @@ def teardown_router(self):
         done here. May return a deferred.
         """
 
+    def get_destination_channel(
+            self, destination_id, message_body=None, message_id=None):
+        """
+        Gets the channel associated with the specified destination.
```
rudigiesler: Can we add here that either `message_body` or `message_id` will be supplied?

junebug/router/from_address.py
```diff
@@ -135,6 +136,11 @@ def setup_router(self):
             self.consume_destination(
                 destination['id'], self.handle_outbound_message)
 
+    def get_destination_channel(
+            self, destination_id, message_body=None, message_id=None):
+        config = self.get_static_config()
+        return succeed(str(config.channel))
```
rudigiesler, line 142: Can we add a test for this method? Since we don't use the from address router in our API tests.

rudigiesler (commented):
Just two small comments :) Looking good

rudigiesler (approved):

PR #166, opened 2018-01-18
Disable channel messages endpoint

erikh360:
... when there is no destination configured on the channel. Router/destination message endpoints should be user in this case.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -277,7 +277,7 @@ class JunebugApi(object):
                 request, 'message submitted', msg, code=http.CREATED))
         else:
             raise ApiUsageError(
-                'This channel has no destinations')
+                'This channel has no "mo_url" or "amqp_queue"')
 
     @app.route(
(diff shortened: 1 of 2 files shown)


junebug/api.py
```diff
@@ -267,20 +267,34 @@ def get_logs(self, request, channel_id):
...
-        returnValue(response(
-            request, 'message submitted', msg, code=http.CREATED))
+        if (channel.has_destination):
+            msg = yield self.send_messsage_on_channel(channel_id, body)
+
+            returnValue(response(
+                request, 'message submitted', msg, code=http.CREATED))
+        else:
+            raise ApiUsageError(
+                'This channel has no destinations')
```
rudigiesler: Maybe we should make this more specific, like mentioning the `mo_url` and `amqp_queue` parameters. `destinations` is not all that clear to users of the API, especially since routers also have the idea of destinations.

rudigiesler (approved):

PR #169, opened 2018-01-22
Fix sending reply from destination

erikh360:


rudigiesler:
@erikh360 Maybe a better solution here is, on the channel, have the current method fetch the related inbound message, and then the rest of the function we abstract out into another function, that takes in the inbound message object to reply to.

Then for replying on the destination endpoints, we fetch the message from the message store ourselves, and pass it to that new second function.

The original function will just load the message, and then call the new second function with that message. 

I hope that makes sense?

rudigiesler:
Then I think there's one more thing that we need to think about here. When we're sending the message using the channel, the channel will be storing it in the message store under the channel ID, not the destination ID, so we won't be able to find the message under the destination.

So we'll have to add it to the message store ourselves under the destination ID.

The other thing we can do here is one of our original plans, or removing the link of a message to a destination / channel, so that we can access messages from anywhere.

erikh360:
Thanks @rudigiesler , I understand the first comment and it makes sense to do it that way.

I'll also store the message under the destination ID.

rudigiesler (approved):

PR #173, opened 2018-07-30
Add destinations to health check

erikh360:


setup.py
```diff
@@ -23,12 +23,12 @@ def read_file(filename):
         # We install pyasn1 first, because setuptools gets confused if it
         # installs pyasn1-modules first.
         'pyasn1',
-        'klein',
+        'klein<=17.10.0',
         'jsonschema',
-        'treq<16.12.0',
+        'treq<=17.8.0',
```
rudigiesler, line 28: This is a `<=`, so it won't actually enforce a minimum version. We should ensure that if we need a later version of treq, then we should require it as a minumum.

erikh360: The minimum version is on Twisted: `Twisted>=17.9.0,<18.0.0`

rudigiesler (approved):

PR #177, opened 2020-04-17
Fix rabbitmq health check

erikh360:


junebug/tests/test_api.py
```diff
@@ -103,20 +103,16 @@ def test_http_error(self):
...
+                    'code': 308,
+                    'message': '308 Permanent Redirect: None',
                     'new_url': '%s/channels/' % self.url,
-                    'type': 'Moved Permanently',
+                    'type': 'Permanent Redirect',
                 }],
             })
-        yield self.assert_response(
-            resp, http.OK,
-            'channels listed', [])
```
rudigiesler, line 119: Is there a reason for removing this assert? This makes sure that the URL we redirected to works.

erikh360: It doesn't actually do the redirect, `[redirect] = resp.history()` failed so I had to remove it. 

The first assert is now on `resp` because `redirect` doesn't exist. @rudigiesler

rudigiesler: Ah, it must be the client that doesn't understand 308 now. 👍

rudigiesler (approved):

PR #178, opened 2020-04-18
Only use message_stats if available

erikh360:
I looked at the response we're getting on QA and `message_stats` isn't always there, in that case we're falling back to `messages_details.rate` which is always there.

### DIFF
--- a/junebug/api.py
+++ b/junebug/api.py
@@ -705,12 +705,10 @@ class JunebugApi(object):
                         if("message_stats" in queue):
                             details['rate'] = \
                                 queue['message_stats']['ack_details']['rate']
-                        else:
-                            details['rate'] = queue['messages_details']['rate']
 
-                        if (details['messages'] > 0 and details['rate'] == 0):
-                            stuck = True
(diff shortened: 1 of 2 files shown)


junebug/api.py
```diff
@@ -699,9 +699,15 @@ def return_queue_results(results):
...
-                            'rate':
-                                queue['message_stats']['ack_details']['rate']
                         }
+
+                        # Use the rate from message_stats if it is there
+                        if("message_stats" in queue):
+                            details['rate'] = \
+                                queue['message_stats']['ack_details']['rate']
+                        else:
+                            details['rate'] = queue['messages_details']['rate']
```
rudigiesler: I don't think that we should be using `messages_details` rate at all, it's very misleading.

I'm taking a wild guess that if a queue hasn't seen activity for a long while (quite possible in QA), then it won't have the `message_stats` key.

If that's the case, I would rather skip checking that queue than rely on `messages_details.rate`.

rudigiesler (approved):

